PR #4, opened 2023-08-25
Patches are MIT by default

MartijnMuijsers:
Quick PR that makes the license text the same everywhere~


Dreeam-qwq:
Thanks a lot!

PR #13, opened 2023-12-04
Fix the sprint glitch which results in inconsistent health.

pog5:
This bug has been in the game since 1.16.4 and has been annoying several of my customers. I would rather not have to make a fork of this every time I need to provision a new server.


Dreeam-qwq:
Thanks to contribute and fix this bug!

PR #30, opened 2024-03-11
Ability to use Virtual Thread for async scheduler

nostalfinals:
This patch add a option (`performance.use_virtual_thread_for_async_scheduler`) to use the new Virtual Thread introduced in JDK 21 for `CraftAsyncScheduler`.
The server jar need to be compiled using JDK 21, and at least Java 21 to run.
This feature will be disabled by default.

## TODO

- [x] Fix compiling issue while shadow plugin doesn't support Java 21 for now -> Fixed by using a shadow fork.
- [x] Test plugins which are using `BukkitScheduler` on this.
- [x] Fix `An expected class  org.bukkit.craftbukkit.v1_20_R3.scheduler.CraftScheduler$4 was not found for preloading: org.bukkit.craftbukkit.v1_20_R3.scheduler.CraftScheduler$4` issue by editing Spigot's preloading code.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,7 +1,8 @@
 plugins {
     java
     `maven-publish`
-    id("com.github.johnrengelman.shadow") version "8.1.1" apply false
+    id("com.github.johnrengelman.shadow") version "8.1.1" apply false // paperweight required this original shadow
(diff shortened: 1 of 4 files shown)


Dreeam-qwq:
Thanks for contributing! merged.

PR #32, opened 2024-03-12
ChatImage Protocol

HaHaWTH:
This pull request adds chat image protocol to Leaf, it works fine during tests.

The client side mod can be found here: https://www.curseforge.com/minecraft/mc-mods/chatimage


HaHaWTH:
0077-Skip-entity-scheduler-when free: When an entity is not being scheduled by any plugins, skip it to get performance optimizaions

PR #55, opened 2024-05-12
[Feature] Configurable bamboo collision

HaHaWTH:
This PR makes bamboo collision configurable.

Disabling collision will fix Geyser players' buggy movement when they close to these blocks,

but all players can pass through bamboos directly(when collision is disabled).

Tested on Geyser (git-master-627c2ba), works fine for me >v<


Dreeam-qwq:
Thanks a lot!!

PR #68, opened 2024-06-14
Reduce worldgen allocations by reusing supplier objects

HaHaWTH:
This change optimizes the way SurfaceRules update their biome supplier,avoiding unnecessary object creations and thus reducing memory allocations
during world generation. The update method now reuses the existingPositionalBiomeGetter object if it's already present, otherwise it
initializes a new one.

Additionally, the tryApply method in SurfaceRules now avoids iterator
allocation by directly accessing the rules list, which further contributes
to reducing garbage collection pressure during world generation.


Dreeam-qwq:
🚀🚀🚀

PR #69, opened 2024-06-14
Optimize Thread Wait Logic in BlockableEventLoop

HaHaWTH:
This PR should fix MC-183518. The current implementation in BlockableEventLoop involves yielding the thread before parking, which is unnecessary and can lead to inefficient thread scheduling and higher CPU usage.

Changes made:
- Replaced the existing thread yielding logic with `LockSupport.parkNanos` to directly park the thread for 2 milliseconds.

Reasoning:
Yielding the thread is redundant if the thread is about to be parked immediately afterward. 

Reference:
- [MC-183518](https://bugs.mojang.com/browse/MC-183518)


Dreeam-qwq:
GO

PR #87, opened 2024-07-21
Fix MC-266334

PadowYT2:
The bug was introduced in 1.20.2+ and was fixed in 1.21

### DIFF
--- a/patches/server/0095-Fix-MC-266334.patch
+++ b/patches/server/0095-Fix-MC-266334.patch
@@ -1,0 +1,27 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: PadowYT2 <me@padow.ru>
+Date: Sun, 21 Jul 2024 15:16:54 +0300
+Subject: [PATCH] Fix MC-266334
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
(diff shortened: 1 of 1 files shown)


patches/server/0095-Fix-MC-266334.patch
```diff
@@ -0,0 +1,27 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: PadowYT2 <me@padow.ru>
+Date: Sun, 21 Jul 2024 15:16:54 +0300
+Subject: [PATCH] Fix MC-266334
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
```
PadowYT2, line 7: Not sure if this should be added, but this was done by git

Dreeam-qwq: Yeah, this is auto generated by git for compatibility

HaHaWTH:
I prefer add a configuration for this fix.
Some servers may use this "bug" to build farms i think

PadowYT2:
> I prefer add a configuration for this fix. Some servers may use this "bug" to build farms i think

This is rather kinda needed, when people build shulker farms, they need the armor stand to **not** break. That's why it was fixed in 1.21 and worked fine in 1.20.1

There are a lot of farms built for 1.20.1 and they don't work on 1.20.2+

F3F5:
឵឵឵ ឵឵឵

Dreeam-qwq:
Thanks!

PR #94, opened 2024-08-04
[Feature] Hide specified item components to clients

TheFloodDragon:
## From the Comment
Controls whether specified component information is sent to clients.
This may break resource packs and mods that rely on this information.
It needs a component type list, incorrect things will not work.
You can fill it with ["custom_data"] to hide components of CUSTOM_DATA.
Also, it can avoid some frequent client animations.
NOTICE: You must know what you're filling in and how it works! It will handle all items!

## WARNING

If you set the value to ["enchantments"], you cannot see any item's enchantments,
also to avoid some frequent client animations,
but the enchantment data is truly in the server.

*So, We only recommend filling it with ["custom_data"]*

**Tests have been done successfully.**


HaHaWTH:
`dev-imports.txt` file change should not be committed i think

TheFloodDragon:
> `dev-imports.txt` file change should not be committed i think

Sorry, My bad, I understand this by mistake. I just saw it in the document.

Dreeam-qwq:
Thanks!

PR #131, opened 2024-10-21
Cleanup protocol code & Further reduce worldgen allocation

HaHaWTH:
This pull request cleanup the ChatImage Protocol comment and unused code.

Avoid lambda & Iterator allocation during worldgen


Dreeam-qwq:
Looks good.

PR #132, opened 2024-10-21
Update README.md

Taiyou06:
removed the "for chinese only" text since afdian accepts paypal and stripe now.
image


Dreeam-qwq:
Thanks to point it out!

PR #133, opened 2024-10-22
Cache blockstate cache

HaHaWTH:
This pull request cached blockstate cache SupportType and Direction.Axis cuz there's no need to call Enum#values() every time using cache


Dreeam-qwq:
Thanks!, goooo!

PR #140, opened 2024-10-27
Further reduce memory footprint of CompoundTag

HaHaWTH:
This pull request further reduces memory footprint of CompoundTag on changes of Paper by interning String keys using Caffeine WeakInterner


Dreeam-qwq:
Lets goooo

PR #141, opened 2024-10-28
Ensure teleport is executed on server thread with Async Tracker on

HaHaWTH:
This pull request removed thread check for MultithreadedTracker by extending TickThread, which could silence thread check errors cuz its safe running on Async Tracker Thread.

https://pastes.dev/IDVAVaXYqW


Dreeam-qwq:
Thanks a lot!

PR #143, opened 2024-10-29
Optimize Entity#distanceToSqr

HaHaWTH:
This pull request optimizes Entity#distanceToSqr call by using Math#fma which is around 1.2x to 4x faster than original method on CPUs support Fused Multiply-Add (FMA) operation,
avoids multiple casting in Entity#distanceTo, using Math#hypot instead of Mojang's Mth#sqrt. Additionally, this patch makes
these methods more able to be inlined by the JIT compiler.

FMA can be enabled via system property.


Dreeam-qwq:
斜斜

PR #149, opened 2024-11-02
Avoid useless deque clear on  LevelTicks#cleanupAfterTick

MachineBreaker:


Dreeam-qwq:
How to replicate this high usage on clear() calls? in what kind of condition (

Some of my test results ( can't show clear() calls of list / array deque have performance issue (
image
image

Taiyou06:
By the time cleanupAfterTick() runs, toRunThisTick is guaranteed to be empty
Therefore, calling clear() on an already empty queue is unnecessary, gonna include this on #217

PR #150, opened 2024-11-03
Better event check for MultithreadedTracker and AsyncLocator

HaHaWTH:
This pull request includes following changes:
1. Added thread check for MultithreadedTracker when re-scheduling events, only schedule events called on TrackerThread to server thread.
2. Use same implementation for AsyncLocator as MultithreadedTracker


Dreeam-qwq:
蟹蟹٩('ω')و

PR #158, opened 2024-11-10
pretty simple patch

Taiyou06:
- optimize getWorld(UUID) by Object2ObjectLinkedOpenHashMap


patches/server/0146-optimize-getWorld-UUID.patch
HaHaWTH, line 1: Remove import here, use full-qualified name( ´◔ ‸◔')

HaHaWTH, line 1: Also this patch should be renamed to ` MultiPaper Optimize getWorld uuid` and add the origin in patch header

Dreeam-qwq:
小恐龙

PR #166, opened 2024-11-15
Cleanup Lithium patches & Faster distanceTo & Optimize worldgen math

HaHaWTH:
This pull request includes following changes:
1. Cleanup useless casting in Lithium equipment tracking
2. Use Math#sqrt directly in Entity#distanceTo method, this will be replaced with internal implementation by JVM, still faster than Mojang's Mth#sqrt since it avoids multiple casting qwp
3. C2ME optimize worldgen (Co-authored by @Taiyou06 )


patches/server/0154-C2ME-Optimize-world-gen-math.patch
```diff
@@ -0,0 +1,87 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: HaHaWTH <102713261+HaHaWTH@users.noreply.github.com>
+Date: Tue, 9 Nov 2077 00:00:00 +0800
+Subject: [PATCH] C2ME: Optimize world gen math
+
```
Dreeam-qwq, line 5: Needs to add header comment about patch source
like https://github.com/Winds-Studio/Leaf/blob/27134f699a8e1c75b6d080d1d6ad7001d3a2aa19/patches/server/0141-C2ME-Reduce-Allocations.patch#L6

if the patch bases on multiple mixins,
then it should be like 
```
...based on the following mixins:
* "path/to/package/MixinA.java"
* "path/to/package/MixinB.java"
....
```

Dreeam-qwq (changes requested):

Dreeam-qwq (approved):

PR #174, opened 2024-11-29
Smooth teleport config

HaHaWTH:
This pull request made smooth teleport not only through bukkit api calls but also through config, requires origin world and target world to have same logical height to work properly.

For example, overworld->nether/world_the_end will not trigger smooth teleport, but overworld->plot(which is backed by overworld) does.

Tested with Multiverse-core and MyWorld plugin, works fine.

Make it as a separate patch to clarify which part is taken from Slice.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,7 +1,7 @@
 plugins {
     java
     `maven-publish`
-    id("io.papermc.paperweight.patcher") version "1.7.4-SNAPSHOT"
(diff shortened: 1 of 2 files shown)


HaHaWTH:
Some notes: variable `i` in PlayerList is used to indicate whether this respawn is a death respawn/dimension switch.

Reference: https://wiki.vg/Protocol#Respawn

Dreeam-qwq (approved):

Dreeam-qwq:
sloped sloped

PR #178, opened 2024-12-01
Move config annotations

HaHaWTH:


patches/server/0049-Linear-region-file-format.patch
```diff
@@ -351,9 +350,7 @@ index 0000000000000000000000000000000000000000..7f95e17544b762ad4674315e53fcff8e
 +        return EnumConfigCategory.MISC.getBaseKeyName() + ".region-format-settings";
 +    }
 +
-+    @DoNotLoad
 +    private static final Logger logger = LogUtils.getLogger();
-+    @DoNotLoad
```
Dreeam-qwq: Do not remove these two annotations, they are necessary for the future config reload udpate

patches/server/0003-Leaf-Config.patch
```diff
@@ -81,24 +88,30 @@ index 0000000000000000000000000000000000000000..f05e1ee109a36c39857f75c296f2ff21
 +            module.onLoaded();
 +
 +            modules.add(module);
++            for (Field field : getAnnotatedStaticFields(clazz, Experimental.class)) {
++                Object obj = field.get(null);
++                if (!(obj instanceof Boolean)) continue;
++                boolean enabled = (Boolean) obj;
++                if (enabled) {
++                    LeafConfig.LOGGER.warn("Experimental feature {} is enabled, please report any bugs you found!", clazz.getSimpleName());
```
Dreeam-qwq: It's better to collect those exp feature names, and warn after the loop, to prevent console spam

Dreeam-qwq (changes requested):

Dreeam-qwq (approved):

PR #190, opened 2024-12-31
Fix player profileresult cache

HaHaWTH:
Make caffeine cache static here, let player ProfileResult cache actually work


Dreeam-qwq:
Happy 2025~

PR #199, opened 2025-01-17
Tick

sfclog:


Dreeam-qwq:
Parallel world ticking or similar concept features will stay at the separate feature branch. They will not be merged to the main branch, since they are experimental and are incompatible with some plugins.

PR #202, opened 2025-01-18
LIGHTNINGLEAF Notice

sh2kcs:
To be honest, I just want to be one of the contributors:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,8 +9,7 @@
 </div>
 
 > [!WARNING]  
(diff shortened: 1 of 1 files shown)


Energobro:
to be honest, you are clown!

Taiyou06:
> to be honest, you are clown!

Dude hasn't done anything wrong, they only added a notice to a non project related thing as a warning which is completely valid. Be better instead of being a troll.

PR #207, opened 2025-01-24
Cleanup SmartSort code

HaHaWTH:
This part can be directly replaced with Integer.getInteger, there's no need to handle manually


Dreeam-qwq:
佬好好休息>~<

PR #208, opened 2025-01-28
Auto resize tracker thread pool

HaHaWTH:
Added a config option `auto-resize` in MultithreadedTracker to enable dynamically adjust thread pool size, overrides max-thread and keepalive.


Dreeam-qwq:
dalao Jump!

PR #217, opened 2025-02-07
Bunch of side perf improvements

Taiyou06:
This PR should add in-dev stuff that was left over in our private test repository. (Tramontane)


Taiyou06:
https://github.com/Winds-Studio/Leaf/pull/217/commits/34c1344e30e2ea4604758398eaa072f4dd443e20

TODO: Seperate the tickets with "not removal delay" outside the array and "with remove delays". So we avoid and predicate tests.
Suggestion by: @MachineBreaker

Taiyou06:
https://github.com/Winds-Studio/Leaf/pull/217/commits/f40f313e22d4b7e5df4b6266a5b9c0bbe1ce7e91
This commit simplifies the SortedArraySet impl by making it O(N) from O(N+1) from paper's implementation, key differences are:
- Avoids writes via writePos check.
- And all of the logic is in a single for loop.

Taiyou06:
https://github.com/Winds-Studio/Leaf/pull/217/commits/2498d5f7110806b9e242fcfb472171429d9389b3
Noteable optimizations are:
- Removed Redundant Checks:
  - Replaced Math.min(storage.length, list.size()) with list.size() since the storage array is always adequately sized.
  - Eliminated null checks for entities within valid indices, trusting the list's integrity.
  
- Loop Splitting:
  - Split the inner loop into four variations based on the presence of except and predicate, reducing conditional checks within each iteration. This avoids redundant evaluations and streamlines the code path for each scenario.
  
- Predicate Handling:
  - Precomputed checkPredicate to determine once whether the predicate needs evaluation, avoiding repeated null checks inside the loop.
Benchmark: 8459ms --> 5375ms (with 3000 villagers)

Dreeam-qwq:
Great work, will cleanup later, also not need to merge new commits from main branch

leaf-server/minecraft-patches/features/0115-Cache-supporting-block-check.patch
```diff
@@ -0,0 +1,63 @@
...
+                 AABB aabb1 = aabb.move(-movement.x, 0.0, -movement.z);
+@@ -1104,6 +1128,15 @@ public abstract class Entity implements SyncedDataHolder, Nameable, EntityAccess
+         }
+     }
+ 
++
++    // Leaf start - Cache supporting block check
++    // Helper method to get cached supporting block state
++    @Nullable
++    public BlockState getCachedSupportingBlock() {
```
Dreeam-qwq: Needs to investigate the usage of `cachedSupportingBlockState ` or how to use it.

PR #218, opened 2025-02-08
Dont reallocate buckets again

HaHaWTH:
Removed bucket reallocation in NearestLivingEntitySensor as buckets have been cleared already


Dreeam-qwq:
天气冷

PR #224, opened 2025-02-11
Thread safety fixes

HaHaWTH:
This pull request fixes some thread safety bugs found in async features.

## Changes

- Fix AsyncMobSpawn
- Fix MultithreadedTracker (ConcurrentLongHashSet)

## To reproduce

1. Set max-players to 50
2. 50+ players join server at the same time
3. See error

## Logs

- https://mclo.gs/DB9lCAZ
- https://mclo.gs/LMtU9Mk

### DIFF
--- a/leaf-server/minecraft-patches/features/0084-Multithreaded-Tracker.patch
+++ b/leaf-server/minecraft-patches/features/0084-Multithreaded-Tracker.patch
@@ -23,6 +23,19 @@ for the case of some NPC plugins which using real entity type, e.g. Citizens.
 But it is still recommending to use those packet based, virtual entity
 based NPC plugins, e.g. ZNPC Plus, Adyeshach, Fancy NPC, etc.
 
+diff --git a/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java b/ca/spottedleaf/moonrise/patches/chunk_system/player/RegionizedPlayerChunkLoader.java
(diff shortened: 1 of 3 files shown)


Dreeam-qwq (approved):

FireInstall:
I'm sorry but using ConcurrentHashMap compleatly defeats the point of FastUtil.
If you are going to create a class with a backing map, it should look something like this: [fastutil-concurrent-wrapper](https://github.com/trivago/fastutil-concurrent-wrapper).

Taiyou06:
It's not ConcurrentHashMap, It's custom made [ConcurrentLongHashSet](https://github.com/Winds-Studio/Leaf/blob/802a83e09571b86cce06c1994f6a4c16b9682504/leaf-server/src/main/java/org/dreeam/leaf/util/map/ConcurrentLongHashSet.java), in theory this would outperform the fastutil-concurrent-wrapper, but feel free to check if there's any benefit of using trivago's solution.

FireInstall:
Ah, I just saw this PR and was shocked it was merged being just a wrapper for ConcurrentHashMap. Didn't check if there were made any changes to the class after that.
Yeah, your solution looks fine to me. Maybe just maybe there could be a case to be made using ReentrantReadWriteLock instead of ReentrantLock, but that's just peanuts in the end.
Alright happy to see this was already addressed. Have a nice day in that case ~

Taiyou06:
Thank you for reaching out your concerns to begin with! if you see anything problematic in the future feel free to let us know!

PR #232, opened 2025-02-19
Add PlayerInventoryOverflowEvent

Dreeam-qwq:
See javadocs comment of PlayerInventoryOverflowEvent class for detail.

- [x] Use handler list to check instead of using explicit enable / isEnabled


Taiyou06 (approved):
Looks clean

HaHaWTH (approved):

PR #235, opened 2025-02-21
Connection.flushQueue optimization

Taiyou06:
Switched from `ConcurrentLinkedQueue` with `ArrayDeque` and for thread safety submits to the netty event loop.


Taiyou06:
Before:
image
After: 
image
(Change is in `Connection.flushQueue`)

Taiyou06:
Tho i also should add that this will increase the amount of work netty executors will do, so it should be advised to give 1-2 additional threads _just in case_

andreasdc:
> Tho i also should add that this will increase the amount of work netty executors will do, so it should be advised to give 1-2 additional threads _just in case_

Is the change for all packets or just for the flushes? If it's for all packet then it's not good.

Taiyou06:
> > Tho i also should add that this will increase the amount of work netty executors will do, so it should be advised to give 1-2 additional threads _just in case_
> 
> Is the change for all packets or just for the flushes? If it's for all packet then it's not good.

Just flushes.

PR #245, opened 2025-03-05
Async playerPacket sending

Taiyou06:
Pretty self explanatory
Before:
image

After:
image

Included options:
image


andreasdc:
Are you sure that makes sense? Won't it collide with flush consolidation? Can you benchmark general CPU usage?

Taiyou06:
> Are you sure that makes sense? Won't it collide with flush consolidation? Can you benchmark general CPU usage?

Netty Flush is another thing, that's just basic ConcurrentLinkedQueue to ArrayDeque and uses netty for thread safety, that's simple and wouldn't cause any issues with plugins.

This on the other hand changes how packets sent which may or may not cause issues with other plugins hence it's toggleable. and cpu consumption wise, i didn't see much of a difference. Which makes sense as this is a singular method that's being off-threaded.

Taiyou06:
Which netty flush is also toggleable now as well, if you saw the commits. 
image

Taiyou06:
Will continue on this after I rework the Async Chunk send for Parallel World Ticking.

PedroMPagani:
Does it not make more sense to make a queue for the packets and jump rhe encoder instead and have a packet encoder for the play state that is threaded? I dont see how this here will trully improve unless packet sending inside tick qctually has a larger cost
Maybe instead have a queue like folia or possibly implement this sort of queue to serialise packets off of the netty threads?

PedroMPagani:
Obviously, fhis is clearly an overhead on the send packet and theres not really a bit benefit having this just be multi threaded, maybe instead flush once inside a tick

Taiyou06:
> I dont see how this here will trully improve unless packet sending inside tick qctually has a larger cost Maybe instead have a queue like folia or possibly implement this sort of queue to serialise packets off of the netty threads?

With player counts upwards of 300+ this does actually become a bottleneck. But your suggestion for a queue based approach could definitely work, I'll fiddle with this a bit to see what i can do 👍 

> Obviously, fhis is clearly an overhead on the send packet and theres not really a bit benefit having this just be multi threaded, maybe instead flush once inside a tick

Yeah lol, MT would slow down due the synchronizations etc.

Taiyou06:
Also you can DM me in Discord (taiyouh), if you wanna share some ideas or code snippets directly 👍

Taiyou06:
- Connection.FlushQueue rewrite patch is removed
- Added AsyncPacketSend with values shown below for customizability;
image

Dreeam-qwq:
🔥🔥🔥

PR #249, opened 2025-03-08
Entity improvements

Taiyou06:
Optimizes bunch of things related to villagers and other mobs, pretty safe optimizations, feel free to test and report any findings.


Taiyou06:
Instead of one big PR, i'll be slowly adding stuff in here to the main directly with benchmarks included.

PR #274, opened 2025-03-31
Allow Non-ASCII character in unquoted string of command argument

hayanesuru:
#273

Client will display an error but still able to send.
It changed decode string of SNBT

test command:
```
/data merge storage ns:test {String:テスト}
/data get storage ns:test String
```


Dreeam-qwq:
Thanks!

PR #303, opened 2025-04-30
optimize async target finding

hayanesuru:
before this pr
```
entitiy tick start
  ...
  entitiy tick ai
entitiy tick end
```

apply this pr and enable async target finding
```
iter all entitities start

entitiy tick start
  ...
  entitiy tick ai partial
entitiy tick end

entitities tick ai partial

entitiy tick start
  ...
  entitiy tick ai partial
entitiy tick end

entitities tick ai partial
...
iter all entitities end
...
block entitiy tick end
entitities tick ai partial
block entitiy tick end
entitities tick ai partial
```

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/config/modules/async/AsyncTargetFinding.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/config/modules/async/AsyncTargetFinding.java
@@ -1,15 +1,9 @@
 
 package org.dreeam.leaf.config.modules.async;
 
-import com.google.common.util.concurrent.ThreadFactoryBuilder;
-import org.dreeam.leaf.async.ai.AsyncGoalExecutor;
 import org.dreeam.leaf.config.ConfigModules;
 import org.dreeam.leaf.config.EnumConfigCategory;
 import org.dreeam.leaf.config.annotations.Experimental;
-
-import java.util.concurrent.ArrayBlockingQueue;
-import java.util.concurrent.ThreadPoolExecutor;
-import java.util.concurrent.TimeUnit;
 
 public class AsyncTargetFinding extends ConfigModules {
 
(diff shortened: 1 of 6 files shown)


leaf-server/src/main/java/org/dreeam/leaf/config/modules/async/AsyncTargetFinding.java
```diff
@@ -20,18 +14,17 @@ public String getBasePath() {
     @Experimental
     public static boolean enabled = false;
     public static boolean alertOther = true;
-    public static boolean searchBlock = false;
+    public static boolean searchBlock = true;
     public static boolean searchEntity = true;
-    public static boolean searchPlayer = false;
-    public static boolean searchPlayerTempt = false;
+    public static int queueSize = 100_000;
```
Taiyou06: This should be specified with the amount of threads available to the executor. (example like how AsyncPathfinding handles it, THREAD_COUNT * 384 type deal.)

Taiyou06: 0 should have the automatic queue limit handled by the thread count * xyz formula, and whenever changed, it should use that value instead.

hayanesuru: > This should be specified with the amount of threads available to the executor. (example like how AsyncPathfinding handles it, THREAD_COUNT * 384 type deal.)

only single thread used

hayanesuru: > 0 should have the automatic queue limit handled by the thread count * xyz formula, and whenever changed, it should use that value instead.

queue size is per world, I think should't multiply thread count again

PR #309, opened 2025-05-04
Raytrace Entity Tracker

HaHaWTH:
Raytrace Entity Tracker uses async ray-tracing to untrack entities players cannot see, which can reduce bandwidth usage significantly, especially in some massive entities in small area situations.  Also it provides a way to guard against Entity ESP hacks.

Experimental feature, more testers are needed( ´◔ ‸◔')


ljys1216:
我很期待此分支，因为我的服务器带宽紧张，急需这个实验性功能

TH2403y:
膜拜！佬！

wg21655:
How can I test this is working correctly? I surrounded an entity but can still see it.

CAJIO:
It looks great. Perhaps it should be added as an experiment?

Zontyk:
> It looks great. Perhaps it should be added as an experiment?

I think so too

PR #317, opened 2025-05-07
async broadcastChanges

Taiyou06:
It's more or less complete but, I'm open to improvements/suggestions/cleanups.


Aides123:
When will it be officially implemented?

PR #322, opened 2025-05-13
Prevent overriding spark flags

Biquaternions:
This should prevent overriding spark flags defined by the user.

## Example on leaf:
image
image

## Example on paper
image
image
image

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/config/LeafConfig.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/config/LeafConfig.java
@@ -7,6 +7,7 @@ import net.minecraft.Util;
 import org.dreeam.leaf.config.modules.misc.SentryDSN;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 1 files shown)


leaf-server/src/main/java/org/dreeam/leaf/config/LeafConfig.java
```diff
@@ -223,17 +223,23 @@ private static List<String> buildSparkExtraConfigs() {
             "config/gale-world-defaults.yml"
         ));
 
+        String existing = System.getProperty("spark.serverconfigs.extra");
+        extraConfigs.addAll(Arrays.asList(existing.split(",")));
```
Dreeam-qwq: Should have a null check for `existing`, same below.

Dreeam-qwq (changes requested):

leaf-server/src/main/java/org/dreeam/leaf/config/LeafConfig.java
Biquaternions: My bad for trying to do it fast, check the next commit

Dreeam-qwq:
Looks good, thanks for your first contribution!

PR #323, opened 2025-05-14
optimize tracker

hayanesuru:


HaHaWTH (commented):
Generally looks good, is there any significant performance difference after switching to synchronized ReferenceOpenHashSet?

PR #326, opened 2025-05-15
Add configurable death item drop knockback

adabugra:
Allow configuration of item drop velocity/knockback on player death


Taiyou06:
Generally looks good to me

PR #335, opened 2025-05-25
Translation of settings

wling-art:
Because the configuration comments seemed incomplete, I tried adding and modifying some of the translations.
Since these comments were added based on my own understanding and I'm not an expert in this area, I really need someone to proofread and verify their accuracy. plz


Dreeam-qwq:
Explanations of config options are on the Leaf docs, and we are still working on it.

PR #349, opened 2025-06-04
Add BlockExplosionHitEvent

Pascalpex:
This event is called when a block is about to execute its `onExplosionHit` method.

You can cancel the event to prevent the block from executing this method.


Dreeam-qwq:
crab crab

Yu-Dino:
- In Spigot, you can only disable and enable wind charge explosions.

- For KnockFFA, Pascalpex needs players to be thrown around but trapdoors not to be triggered.

- Pascalpex must prevent the effects on blocks without disabling the entire explosion.

- This event allows you to modify the effects on blocks during an explosion (without preventing the explosion itself).

PR #355, opened 2025-06-07
Fix NPE in entity tracker when chunk data is null

unitdevgcc:


Taiyou06:
Do you know how we can replicate the issue? Because moonrise shouldn't have any chunk data being null. If it's from moonrise, it should be fixed properly on the chunk system instead of on the tracker.

PR #357, opened 2025-06-08
optimize random tick

hayanesuru:


DeadByDays:
crash https://mclo.gs/WAwOK2T

PR #380, opened 2025-06-26
Optimize BlockEntityTickersList#removeAllByIndex

unitdevgcc:
The previous implementation of removeAllByIndex iterated through the list and moved elements one by one. This is inefficient for sparse removals.

    This commit refactors the method to use System.arraycopy to move contiguous blocks of elements, avoiding one-by-one copies. This resolves the TODO and should provide a noticeable performance improvement when removing multiple block entity tickers at once.

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/config/LeafGlobalConfig.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/config/LeafGlobalConfig.java
@@ -141,17 +141,41 @@ public class LeafGlobalConfig {
 
     public Double getDouble(String path) {
         String value = configFile.getString(path, null);
-        return value == null ? null : Double.parseDouble(value); // TODO: Need to check whether need to handle NFE correctly
+        if (value == null) {
+            return null;
+        }
+        try {
+            return Double.parseDouble(value);
+        } catch (NumberFormatException e) {
+            org.leavesmc.leaves.LeavesLogger.getLogger().warn(path + " is not a valid number! Please check your configuration.", e);
+            return null;
+        }
     }
 
     public Integer getInt(String path) {
         String value = configFile.getString(path, null);
-        return value == null ? null : Integer.parseInt(value); // TODO: Need to check whether need to handle NFE correctly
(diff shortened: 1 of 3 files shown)


leaf-server/src/main/java/org/dreeam/leaf/config/LeafGlobalConfig.java
```diff
@@ -141,17 +141,41 @@ public String getString(String path) {
...
     public Long getLong(String path) {
         String value = configFile.getString(path, null);
-        return value == null ? null : Long.parseLong(value); // TODO: Need to check whether need to handle NFE correctly
+        if (value == null) {
+            return null;
+        }
+        try {
+            return Long.parseLong(value);
+        } catch (NumberFormatException e) {
+            org.leavesmc.leaves.LeavesLogger.getLogger().warn(path + " is not a valid number! Please check your configuration.", e);
```
Dreeam-qwq, line 176: Use `LeafConfig#LOGGER`, same for other places

leaf-server/src/main/java/org/dreeam/leaf/config/modules/misc/ConnectionMessage.java
```diff
@@ -38,13 +38,28 @@ public void onLoaded() {
...
+            needsSave = true;
+        }
+        if (quitMessage.contains("%")) {
+            quitMessage = quitMessage
+                .replace("%player_name%", "<player_name>")
+                .replace("%player_displayname%", "<player_displayname>");
+            config.set(getBasePath() + ".quit.message", quitMessage);
+            needsSave = true;
+        }
+        if (needsSave) {
```
Dreeam-qwq, line 57: No need to save here, config will save to file in `loadAfterBootstrap` (after init)

Dreeam-qwq (changes requested):

Dreeam-qwq:
Added in https://github.com/Winds-Studio/Leaf/commit/4d463665940cf5d4b145b4e71241645446309abf

PR #390, opened 2025-07-05
Refactor async entity tracker

hayanesuru:


HaHaWTH (approved):

leaf-server/src/main/java/org/dreeam/leaf/config/modules/async/MultithreadedTracker.java
```diff
@@ -48,7 +49,7 @@ public void onLoaded() {
         if (asyncEntityTrackerMaxThreads < 0)
             asyncEntityTrackerMaxThreads = Math.max(Runtime.getRuntime().availableProcessors() + asyncEntityTrackerMaxThreads, 1);
         else if (asyncEntityTrackerMaxThreads == 0)
-            asyncEntityTrackerMaxThreads = Math.max(Runtime.getRuntime().availableProcessors() / 4, 1);
+            asyncEntityTrackerMaxThreads = 1;
```
HaHaWTH: This should be reverted, let it dynamically scale based on cpu cores

Taiyou06 (approved):
Overall looks good IMHO haven't saw any logic errors myself.

PR #391, opened 2025-07-05
Dev/tracker improvements

Taiyou06:
Replaces thread pool approach with short lived VT threads that are created per tick

That way people won't bother with thread counts and queue being an issue. as each task will be handled at one point on the virtual thread.

(needs heavy testing and reviews)


hayanesuru:
> Virtual threads are suitable for running tasks that spend most of the time blocked, often waiting for I/O operations to complete. However, they aren't intended for long-running CPU-intensive operations.

https://openjdk.org/jeps/444

Taiyou06:
> > Virtual threads are suitable for running tasks that spend most of the time blocked, often waiting for I/O operations to complete. However, they aren't intended for long-running CPU-intensive operations.
> 
> https://openjdk.org/jeps/444

These threads won't be long running, they'll be created every tick for select amount of tasks that are tracker needs then gets cleared.

Taiyou06:
haya rewrote the entire tracker here: https://github.com/Winds-Studio/Leaf/pull/390 so this pr is more or less pointless now

PR #393, opened 2025-07-07
fix TRACKED entities not count

### LINKED ISSUE #392: Mobs build up around the player in excess of the mob cap when logging in and out every few seconds
### Checklist

- [x] I have searched the existing issues and the documentation
- [ ] I ticked all the boxes without actually reading them
- [x] I am using latest Leaf build from https://github.com/Winds-Studio/Leaf/releases
- [x] I tested and couldn't replicate this issue on [latest Purpur build](https://purpurmc.org/)

### What behaviour is observed?

Okay, so when testing I found that there was a blowout in entities that just kept building far in excess of the mob cap. A few of the bots I was testing with had issues and were connecting and then disconnecting every few seconds, and the mobs were only building up around those players specifically to in some cases thousands of entities within spawning range.

If I disable async-mob-spawning then the number of entities near those rapidly reconnecting players would stop _increasing_ on the next restart, but the already spawned mobs would continue to persist.

So while it would seem that async-mob-spawning is perhaps "not noticing" the mobs already around the player when they log in, it would also seem that there is at least some relating to mob despawning as well as I would have expected in the few seconds that the player is offline, that all non-persistent mobs previously around them would instantly despawn and it doesn't seem like this is happening.

I went looking for related settings, and found that if I lower delay-chunk-unloads-by in the Paper config from 10s to 1s then the problem goes away (ie. presumably in the few seconds the player is offline all of the nearby mobs are correctly hard despawning prior to chunk unload) but with the longer unload delay of 10s, it seems they might not actually be despawning despite the fact that the player isn't in the world for several seconds.

Could be a stretch, but that's the observations that I've made while testing. Is it possible that they're not hard despawning during the delayed chunk unload like they are supposed to which allows them to persist until the player logs back in, after which those mobs are just not being counted for some reason by the async-mob-spawning code so that the number of mobs around the player can then increase without limit?

### What is the expected behavior?

Mobs around the player should respect the per-player mob cap.

### How to reproduce?

This is how I've been able to reproduce it.
- enable async-mob-spawning
- log in my player to observe
- log in 30 bots for testing (3 of which for some reason have been doing the constant login/logout cycling every few seconds for an unrelated reason in SoulFire)
- watch the mob count increase out of control 
 
I stopped one time when the total mob count hit 16,500 when it should be more like 7,000 and stopped at 10,000 for the provided spark link.

Also confirmed with the paper entity list command that it's not just Spark misreporting the numbers (see the attached server log).

### Spark profiler link

https://spark.lucko.me/irphc6trUv

### Error log (if applicable)

https://mclo.gs/kcjG4wc

hayanesuru:
Closes #392


CC-Odin:
Confirmed. #392 appears to be resolved

PR #410, opened 2025-07-20
Config v4

Taiyou06:
- [x] Auto migrator
- [x] Backup system
- [ ] Proper spark folder handling
- [ ] Rewritten config descriptions


Dreeam-qwq:
Hey Taiyou! Thanks for this suggestion and for contributing to this. However, I need to close this PR.
I finished the config section of the Leaf docs, see [Leaf-website#38](https://github.com/Winds-Studio/Leaf-website/pull/38) and [Leaf-website#39](https://github.com/Winds-Studio/Leaf-website/pull/39). And I prefer to keep less description in the config to make it easy to read, and move the full description to the docs. Another reason is that the config can not include some formats of content.

Also, I think it's better to keep one config file if we can keep all option names and structures organized and clear, and one Leaf config is better for maintenance as well.

PR #413, opened 2025-07-21
Inlined predicate testing for collisions

Taiyou06:
- Re-ordereed logic on isAlive
- Better collection on collidableExemptions
- Inlined predicate testing on EntitySelector.pushable
- Switched license to LGPL-3.0

Looking for any broken logic.


Taiyou06:
Patch will be dropped due the amount of @Overrides and doing manual instanceof check on each entity type would be slower than before.

Taiyou06:
Will port the safer optimizations to the branch tho.

PR #414, opened 2025-07-21
Updating Blake2 hash function to Blake3

GideonWhite1029:
The code from this repository was used for the reference: https://github.com/bcgit/bc-java/blob/main/core/src/main/java/org/bouncycastle/crypto/digests/Blake3Digest.java


HaHaWTH:
Is there any performance benchmark for this BLAKE3 implementation comparing to previous BLAKE2b?

HaHaWTH (commented):
Generally good, but need some benchmarks to make sure it's faster than original blake2b impl. Leave the old implementation as a compatibility option is better i assume

GideonWhite1029:
изображение
изображение
изображение
изображение

There are places in the current Blake3 implementation that need to be optimized, and at the moment Blake2b will be faster compared to Blake3. Alternatively, you can leave both variations of the hash function and select the version through the configuration file.

leaf-server/src/main/java/su/plo/matter/WorldgenCryptoRandom.java
```diff
@@ -24,6 +25,16 @@ public class WorldgenCryptoRandom extends WorldgenRandom {
     private final long[] message = new long[16];
     private final long[] cachedInternalState = new long[16];
 
+    private static long[] hashingWorldSeed(long[] seed) {
```
Dreeam-qwq: Should we move hashing algo type to config, and use interface-impl for 2b and 3?

cc @HaHaWTH

HaHaWTH: Yes? That would be a cleaner implementation, anyway now looks good except this.

HaHaWTH (commented):
Will run some performance benchmarks and validations later

GideonWhite1029:
@Dreeam-qwq Consider the current changes in the implementation of Blake3, as well as evaluate the purity of the code?

PR #418, opened 2025-07-22
Optimize async pathfinding

Taiyou06:
- Fast-path state checks and double-checked locking to reduce synchronization overhead
- Pre-sized collections and memory cleanup to minimize allocations
~~- Batch processing limits (32 tasks) to prevent main thread blocking~~
- Single-position comparison optimization for common pathfinding cases

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
@@ -20,6 +20,8 @@ public class AsyncPath extends Path {
 
     private volatile boolean fastPath = false;
 
+    private volatile boolean processingFinished = false;
+
     /**
      * marks whether this async path has been processed
      */
@@ -92,6 +94,12 @@ public class AsyncPath extends Path {
             return;
         }
 
+        boolean finished = this.processingFinished;
+        if (finished) {
+            runnable.run();
+            return;
+        }
+
         synchronized (this) {
             if (isProcessed()) {
                 runnable.run();
@@ -124,13 +132,18 @@ public class AsyncPath extends Path {
      * starts processing this path
      */
     public void process() {
-        // Double-checked locking pattern for better performance
         if (this.processState == PathProcessState.COMPLETED) {
             return;
         }
 
+        // Can we also check for PROCESSING state here?
+        PathProcessState currentState = this.processState;
+        if (currentState == PathProcessState.COMPLETED || currentState == PathProcessState.PROCESSING) {
+            return;
+        }
+
         synchronized (this) {
-            // Check again after acquiring lock
+            // Double-check after acquiring lock
             if (this.processState != PathProcessState.WAITING) {
                 return;
             }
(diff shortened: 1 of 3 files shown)


leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -103,40 +112,59 @@ public boolean hasSameProcessingPositions(final Set<BlockPos> positions) {
...
 
     /**
      * starts processing this path
      */
-    public synchronized void process() {
-        if (this.processState == PathProcessState.COMPLETED ||
-            this.processState == PathProcessState.PROCESSING) {
+    public void process() {
+        // Double-checked locking pattern for better performance
+        if (this.processState == PathProcessState.COMPLETED) {
```
MartijnMuijsers: Can this also be `!= PathProcessState.WAITING`?

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -103,40 +112,59 @@ public boolean hasSameProcessingPositions(final Set<BlockPos> positions) {
...
+            this.canReach = bestPath.canReach();
 
-        processState = PathProcessState.COMPLETED;
+            processState = PathProcessState.COMPLETED;
 
-        for (Runnable runnable : this.postProcessing) {
-            runnable.run();
-        } // Run tasks after processing
+            // Process callbacks
+            for (Runnable runnable : this.postProcessing) {
```
MartijnMuijsers: You can run these after `synchronized` since nothing can be added to them anymore (since adding only happens within another `synchronized` block that checks completion first).

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -78,17 +80,24 @@ public AsyncPath(@NotNull List<Node> emptyNodeList, @NotNull Set<BlockPos> posit
...
         if (isProcessed()) {
             runnable.run();
-        } else {
-            this.postProcessing.add(runnable);
+            return;
+        }
+
+        synchronized (this) {
+            if (isProcessed()) {
+                runnable.run();
```
MartijnMuijsers: You can store in a local variable whether processing was finished, and then after the `synchronized` block run the runnable if it was (and add to `postProcessing` within `synchronized` if it wasn't)

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPathProcessor.java
```diff
@@ -106,27 +106,24 @@ private static BlockingQueue<Runnable> getQueueImpl() {
...
-                                pendingTask.run();
-                            }
-                        }
-                        rejectedTask.run();
-                    }
-                    case CALLER_RUNS -> rejectedTask.run();
+
+            switch (AsyncPathfinding.asyncPathfindingRejectPolicy) {
+                case FLUSH_ALL -> {
+                    // Process in batches instead of all at once to avoid blocking main thread
```
MartijnMuijsers: I don't understand the context of this code, so I'm not sure if this has any bad effects~

MartijnMuijsers (approved):

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPathProcessor.java
HaHaWTH, line 1: I'm concerned about batching pathfinding tasks, wouldn't it cause small lag spikes instead of a large one? I assume the first one is more annoying(Maybe leave a config option for this?). Also after batching process the config name should be changed.

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPathProcessor.java
```diff
@@ -106,27 +106,24 @@ private static BlockingQueue<Runnable> getQueueImpl() {
...
+                    batch.forEach(Runnable::run);
+                    rejectedTask.run();
                 }
+                case CALLER_RUNS -> rejectedTask.run();
             }
 
-            if (System.currentTimeMillis() - lastWarnMillis > 30000L) {
-                LOGGER.warn("Async pathfinding processor is busy! Pathfinding tasks will be treated as policy defined in config. Increasing max-threads in Leaf config may help.");
-                lastWarnMillis = System.currentTimeMillis();
+            long currentTime = System.currentTimeMillis();
```
MartijnMuijsers: Also just a nitpick: `currentTimeMillis` is not monotonic, `nanoTime` is. In other words, `currentTimeMillis` may cause code to misunderstand how much time elapsed when the system clock is changed, whereas for `nanoTime` it would work fine.

HaHaWTH (approved):

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -78,17 +82,30 @@ public AsyncPath(@NotNull List<Node> emptyNodeList, @NotNull Set<BlockPos> posit
...
+            } else {
+                this.postProcessing.add(runnable);
+            }
         }
```
MartijnMuijsers: I think you misunderstood my previous comment, I meant this:

```suggestion
        boolean shouldRun = false;
        synchronized (this) {
            if (isProcessed()) {
                shouldRun = true;
            } else {
                this.postProcessing.add(runnable);
            }
        }
        if (shouldRun) {
            runnable.run();
        }
```

MartijnMuijsers: This pulls the runnable out of the critical section

PR #419, opened 2025-07-22
fix async tracker bundle packet too large

hayanesuru:
restore to vanilla behavior


Taiyou06:
Like i said in discord as well, why not just have a smaller value? or even better configurable value that starts at some arbitrary small value. As bundling has quite nice benefits as well for the clients with poor connections.

hayanesuru:
https://minecraft.wiki/w/Java_Edition_protocol/Packets#Bundle_Delimiter

they are same, bundle packet is add two delimiter packets to the bundle, not help to clientside networking

Taiyou06:
Ah gotcha

Taiyou06 (approved):
LGTM

PR #423, opened 2025-07-22
Simplified AsyncPath

Biquaternions:
I originally made this change for a private fork I have called Aurora, in which has been working perfectly so far.
Still I haven't tested enough to say is safe to use with Leaf's version of the petal patch.

This has the following improvements:
1. Uses ConcurrentLinkedQueue, which doesn't require to synchronize the main thread to schedule post-processing tasks
2. Simplifies the logic for processing, if the main thread requests process before it's completed, the time to wait for the synchronized block to be released should be equivalent to just running the path on the main thread.
3. Deprecates PathProcessState since only two states are actually relevant for this implementation.
4. Prevents unnecessary scheduling if already running on the main thread
5. Eliminates one unnecessary lambda allocation in AsyncPathProcessor (in Aurora I completely removed the awaitProcessing function, which makes so a single lambda is allocated in total and only if the async path is not ready, if it is, it just runs the function).

As mentioned before, this has been working perfectly for a couple of months now in my fork, but I haven't tested it more than an hour in Leaf, feel free to do additional testing before merging.
I also noticed the patch changed compared to the first time I saw it, if those changes are still required for some internal behavior in Leaf and you decide to merge this, then feel free to edit this PR accordingly.

Feel free to make the change mentioned in 5 as well, I want to avoid editing patches to avoid possible merge conflicts.


MartijnMuijsers:
Thank you for the PR! :)

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -82,35 +85,17 @@ public AsyncPath(@NotNull List<Node> emptyNodeList, @NotNull Set<BlockPos> posit
...
-            } else {
-                this.postProcessing.add(runnable);
-            }
-        }
-
-        if (shouldRun) {
+        if (this.ready) {
             runnable.run();
+        } else {
+            this.postProcessing.offer(runnable);
```
MartijnMuijsers, line 95: This is slightly incorrect: `ready` may return false, after which another thread sets `ready` to true and empties the queue, and then this thread adds the runnable which will never get executed.

MartijnMuijsers: There's many different ways to address it.

Here's a possible way (not the prettiest, but definitely the easiest):

* Extract the following code from `process` into a new method called `runAllPostProcessing()`:
```
while (!this.postProcessing.isEmpty()) { // Only one thread can reach here, it should never poll null
    if (isTickThread) {
        this.postProcessing.poll().run();
    } else {
        MinecraftServer.getServer().scheduleOnMain(this.postProcessing.poll());
    }
}
```

And then modify it to allow it to be run by multiple threads:
```
Runnable runnable;
while ((runnable = this.postProcessing.poll()) != null) {
    if (isTickThread) {
        runnable.run();
    } else {
        MinecraftServer.getServer().scheduleOnMain(runnable);
    }
}
```

And then, after `this.postProcessing.offer(runnable);` add a call to this method:

```
public void postProcessing(@NotNull Runnable runnable) {
    if (this.ready) {
        runnable.run();
    } else {
        this.postProcessing.offer(runnable);
        if (this.ready) {
            this.runAllPostProcessing();
        }
    }
}
```

The chance of `postProcessing` being run by multiple threads is very small. It simply replaces the case where otherwise a runnable would be skipped.

Biquaternions: I've been trying a couple possible fixes but I think yours is the best and simplest one.
I'm pushing the change, lmk if it's ok!

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -78,211 +82,194 @@ public AsyncPath(@NotNull List<Node> emptyNodeList, @NotNull Set<BlockPos> posit
...
-    public synchronized void postProcessing(@NotNull Runnable runnable) {
-        if (isProcessed()) {
+    public final void schedulePostProcessing(@NotNull Runnable runnable) {
+        if (this.ready) {
             runnable.run();
         } else {
-            this.postProcessing.add(runnable);
+            this.postProcessing.offer(runnable);
+            if (this.ready) {
+                this.runAllPostProcessing(true);
```
MartijnMuijsers, line 97: Is `isTickThread` always true here because `schedulePostProcessing` is only called on a tick thread because `awaitProcessing` is only called on a tick thread?

Since this situation is really unlikely anyway, adding a tick thread check here has no impact on performance. Just in case someone decides to call `awaitProcessing` on a non-tick thread later. They're public methods without any documentation that they should only be called on the main thread after all.

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -78,211 +82,194 @@ public AsyncPath(@NotNull List<Node> emptyNodeList, @NotNull Set<BlockPos> posit
...
      */
-    public synchronized void process() {
-        if (this.processState == PathProcessState.COMPLETED ||
-            this.processState == PathProcessState.PROCESSING) {
-            return;
+    public final void process() {
+        if (this.ready) return;
+
+        synchronized (this) {
+            if (this.ready) return; // In the worst case, the main thread only waits until any async thread is done and returns immediately
```
MartijnMuijsers, line 129: Question: does the main thread run `process()` when it _needs_ the path to be done? So, just purely by necessity, the main thread always has to wait at this point while the async thread is not done doing the processing?

MartijnMuijsers: Because I assumed `awaitProcessing` would be called when the main thread actually needs the result.

If the main thread doesn't necessarily need the result when calling `process` yet, we can make an easy improvement here that avoids any waiting:

Replace `synchronized (this)` by `if (!hasStarted.get() && hasStarted.getAndSet(true) == false) {`
and add a new `AtomicBoolean hasStarted = new AtomicBoolean();` to this class.

The `synchronized` is then not necessary at all anymore as long as any thread that accesses any of the content checks `this.ready == true` at some time before that.

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
Biquaternions: Yes, it is always true because it is only called by the main thread.
Personally, if the concern is a plugin calling that method as API, I would instead put an async catcher there to encourage proper development practices 
However, that is my personal opinion, the method can very well just check if it's tick-thread as you mentioned
Let me know what approach convinces you more

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
Biquaternions: Actually, the first implementation I made for Aurora used an atomic boolean with compareAndSet to never wait and always return.
It works perfectly 99% of the time, where the 1% are race conditions where all the path functions overriden below can throw an NPE.
It is possible to check for null inside the patch, but that will cause to completely skip some postProcessing tasks to ever be scheduled and will still cause an NPE in Pathfindinder -> findPath

MartijnMuijsers: This answer makes me a bit scared about the potential cascade of hard-to-debug-from-a-live-server-stack-trace issues from NPEs being thrown within these functions. But your patch doesn't make these better or worse, so that needs to be addressed separately.

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
MartijnMuijsers: Okay, got it! It's fine like this~ just a thing I noticed

MartijnMuijsers (approved):

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
Biquaternions: Reading this again I think I forgot to answer the initial question.
Yes, the main thread calls process when it needs the result.
However, the only function that can result in a NPE if returning too early is getTarget()

The synchronized block exists to guarantee that getTarget always exists when it needs to be.
Also, the NPE is the least of the problems, if the synchronized block is removed all the "getNode..." methods at the end of the class will fail to return a valid node, since the "this.nodes" may not be filled yet, which might not throw an NPE but will definitely cause unknown behavior.

PR #425, opened 2025-07-23
Custom NonNullList

Taiyou06:
Replaces NonNullList's ArrayLists with custom one based on fastutil.ObjectArrayList


leaf-server/src/main/java/org/dreeam/leaf/util/list/OptimizedNonNullListArrayList.java
```diff
@@ -0,0 +1,428 @@
...
+    private static final long serialVersionUID = 1L;
+
+    private static final int DEFAULT_INITIAL_CAPACITY = 10;
+    private static final int MAX_ARRAY_SIZE = Integer.MAX_VALUE - 8; // JVM limit
+    private static final int GROWTH_SHIFT = 1; // equivalent to /2 but faster
+    private static final int CAPACITY_THRESHOLD = 1 << 30;
+
+    private static final int HAS_DEFAULT_FLAG = 2;
+
+    /** Packed flags for various boolean states */
```
MartijnMuijsers, line 35: Currently only used with `HAS_DEFAULT_FLAG`, we could have a `HASH_VALID_FLAG = 1` instead of `hashValid` too to use slightly less memory.

leaf-server/src/main/java/org/dreeam/leaf/util/list/OptimizedNonNullListArrayList.java
```diff
@@ -0,0 +1,428 @@
...
+                // For larger arrays, use more conservative growth to avoid OOM
+                newCapacity = oldCapacity + (oldCapacity >>> 2); // 1.25x growth
+            }
+
+            newCapacity = Math.max(newCapacity, minCapacity);
+
+            if (newCapacity > MAX_ARRAY_SIZE) {
+                newCapacity = hugeCapacity(minCapacity);
+            }
+            if (newCapacity < 1024 && !isPowerOfTwo(newCapacity)) {
```
MartijnMuijsers:
```suggestion
            if (newCapacity < 1024) {
```

Seems not needed to check `isPowerOfTwo` when `nextPowerOfTwo` will just return the input in that case.

leaf-server/src/main/java/org/dreeam/leaf/util/list/OptimizedNonNullListArrayList.java
```diff
@@ -0,0 +1,428 @@
...
+        }
+
+        for (; i < length; i++) {
+            h = 31 * h + array[i].hashCode();
+        }
+
+        cachedHashCode = h;
+        hashValid = true;
+        return h;
+    }
```
MartijnMuijsers:
```suggestion
    @Override
    public int hashCode() {
        if (hashValid) return cachedHashCode;
        
        final Object[] array = this.a;
        final int length = this.size;
        int h;
        int i = 0;
        if (length >=4) {

            int h1 = 1;
            int h2 = 3;
            int h3 = 7;
            int h4 = 11;

            for (; i < length - 3; i += 4) {
                h1 = 31 * h1 + array[i].hashCode();
                h2 = 31 * h2 + array[i + 1].hashCode();
                h3 = 31 * h3 + array[i + 2].hashCode();
                h4 = 31 * h4 + array[i + 3].hashCode();
            }
            h = h1 + h2 + h3 + h4;

        } else {
            h = 1;
        }

        for (; i < length; i++) {
            h = 31 * h + array[i].hashCode();
        }

        cachedHashCode = h;
        hashValid = true;
        return h;
    }
```

Make it even faster by making it so the hashcode additions can be pipelined in parallel.

MartijnMuijsers (approved):

leaf-server/src/main/java/org/dreeam/leaf/util/list/OptimizedNonNullListArrayList.java
```diff
@@ -0,0 +1,442 @@
...
+
+        for (int i = Math.min(size - 1, 3); i >= 0; i--) {
+            if (k.equals(array[i])) return i;
+        }
+
+        return -1;
+    }
+
+    @Override
+    public int hashCode() {
```
MartijnMuijsers, line 294:
```suggestion
    public final int hashCode() { // Must be final to be used in equals(..)
```

PR #427, opened 2025-07-23
optimize MpmcQueue memory layout

hayanesuru:
preventing false sharing


Taiyou06 (approved):
Wish @Contended was accessible

PR #446, opened 2025-08-06
get craftDelegate on copying ItemStack

JWJUN233233:
some plugins like Slimefun may inherit ItemStack and override ItemStack#clone() and return inheritance of ItemStack. It may cause ClassCastException on copying the ItemStack again


HaHaWTH:
According to Bukkit API specification, all interfaces are not designed to be implemented by any plugins, please report to Slimefun.

JWJUN233233:
should i close this issue?

HaHaWTH:
IMG

From Paper javadoc, plugins should never extending or implementing API classes unless with explicit indications.

PR #449, opened 2025-08-07
Bump netty to 4.2.x

Dreeam-qwq:
Tested on my side that all work good, plugin that depends to protocollib can work without issue.
Also for reference: https://github.com/netty/netty/wiki/Netty-4.2-Migration-Guide

Added a system property `-DLeaf.native-transport-type=TYPE` for selecting the network I/O model (Available type: `io_uring`, `epoll`, `nio`). Only affect if `use-native-transport` in server.properties is enabled.

## How to test? (Linux only)
Enable `use-native-transport` in server.properties. And add flag below into your startup flag
```
-DLeaf.native-transport-type=io_uring -Dio.netty.allocator.type=adaptive
```
To enable io_uring transport type and netty's new adaptive allocator.


HaHaWTH (approved):

Dreeam-qwq:
Will remove draft status until Netty 4.2.4 is released

Dreeam-qwq:
Removed draft since netty released 4.2.4.

PR #450, opened 2025-08-08
Drop "Cache random tick block status"

hayanesuru:


Dreeam-qwq (approved):
Thanks to point it out =w=

PR #462, opened 2025-08-14
optimize entity attribute

hayanesuru:
refactor and remove synchronization primitives, move to main thread


Taiyou06:
This needs proper testing for plugins that use attributes like Mythicmobs etc. anyone reading this, feel free to test this through actions build.

PR #465, opened 2025-08-15
Replace GoalSelector#availableGoals with optimized collection

hayanesuru:
faster iteration and sorted


Taiyou06:
is there a spark report with this patch? would be good to see before approving

Taiyou06:
I did check it myself, there's a 2-3mspt drop from before

Taiyou06 (approved):

PR #478, opened 2025-08-25
No collision for spawner-spawned mobs

pixl8dev:
This patch introduces a new setting that enables the disabling of collisions (pushing) for entities spawned from spawners. This feature proves particularly useful when dealing with servers that have many spawner mob farms.


Dreeam-qwq (changes requested):
For the option, I think it's better to use `enable-mob-collision-from-spawner` or `disable-mob-collision-from-spawner` here?

purpurcof:
Any information?

PR #484, opened 2025-08-30
optimize setTicksFrozen

hayanesuru:


Dreeam-qwq:
For the patch name, does `Only update frozen ticks if changed` sound better?

Dreeam-qwq (approved):

PR #493, opened 2025-09-08
Parallel World Ticking API for monitoring tools

Biquaternions:
This just provides an API to retrieve MSPT from each world.
The patch was adapted from [my fish 🐟 fork](https://github.com/Biquaternions/Fish/blob/ver/1.21.8/fish-api/paper-patches/features/0002-Fish-Parallel-World-Ticking-API.patch) and should help any plugin developer to make their own monitoring tools that track mspt per-world.
The API uses the same naming as their equivalent global mspt functions in `Bukkit` class.


Dreeam-qwq (approved):

Dreeam-qwq:
Thank you!

PR #499, opened 2025-09-11
fix(replay): run ServerPhotographer player ops on main thread

wling-art:
Adjust ServerPhotographer to schedule player placement, teleport, and
removal on the main thread when parallel ticking is enabled.

The change was based on TickThread error logs. The related issue did not
include concrete reproduction steps and the problem is hard to reproduce,
so the fix was derived from the stack trace.

Link: https://github.com/Winds-Studio/Leaf/issues/496


Dreeam-qwq (approved):

Dreeam-qwq:
斜斜

PR #332, opened 2025-05-21
Reduce optimiseRandomTick new BlockPos instance create

wling-art:
Reduce optimiseRandomTick new BlockPos instance creation

This patch optimizes the optimiseRandomTick method in ServerLevel by reusing a single MutableBlockPos instance per thread instead of creating a new immutable BlockPos object on every random tick.

- Introduces a ThreadLocal<BlockPos.MutableBlockPos> cache to avoid excessive object allocations.
- Updates the mutable position with .set(...) and converts it to an immutable BlockPos only when passing to randomTick.
- This reduces memory pressure and potential GC overhead during random block ticking.
- Ensures thread safety by using ThreadLocal to avoid concurrent modification issues.

PR #505, opened 2025-09-20
Reduce optimiseRandomTick new BlockPos instance create

wling-art:
Reduce optimiseRandomTick new BlockPos instance creation

This patch optimizes the optimiseRandomTick method in ServerLevel by reusing a single MutableBlockPos instance per thread instead of creating a new immutable BlockPos object on every random tick.

- Introduces a ThreadLocal<BlockPos.MutableBlockPos> cache to avoid excessive object allocations.
- Updates the mutable position with .set(...) and converts it to an immutable BlockPos only when passing to randomTick.
- This reduces memory pressure and potential GC overhead during random block ticking.
- Ensures thread safety by using ThreadLocal to avoid concurrent modification issues.

Original PR: https://github.com/Winds-Studio/Leaf/pull/332


Dreeam-qwq (approved):
Gooood

PR #512, opened 2025-10-01
fix collision shape cache

hayanesuru:
fix shift key collision on powder snow and scaffolding block
likely not affect client side. might fix for NPC


Dreeam-qwq (approved):
Gooood

PR #534, opened 2025-10-18
Fix MC-299642

HaHaWTH:
Mojira link: https://bugs.mojang.com/browse/MC/issues/MC-299642

Minecraft uses 3D movement control for turtles to move in water, which makes them fall even NoGravity tag is set to true.

This direct modification of deltaMovement bypasses the standard move logic where the NoGravity tag is checked and applied.


Dreeam-qwq (approved):
跳起来!

PR #538, opened 2025-10-21
configurable mob goal update interval

hayanesuru:


Dreeam-qwq (approved):
Gooood

leaf-server/src/main/java/org/dreeam/leaf/config/modules/opt/EntityGoal.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public static int chanceTarget = -1; // only all <= 10
+    public static int chanceFollowParent = -1;
+    public static int chanceAvoidEntity = -1;
+    public static int chanceTempt = -1;
+    public static int chanceEndermanLookForPlayer = -1;
+
+    @Override
+    public void onLoaded() {
+        String path = getBasePath() + ".chance";
```
Dreeam-qwq: Whether better to rename `chance` to `tick-chance`, for better semantics.

PR #543, opened 2025-10-21
scan for deprecated config modules

HaHaWTH:
Add support for scanning deprecated features and warning about deprecation


Dreeam-qwq (approved):
slope slope

PR #551, opened 2025-11-01
Fix configurable username validation

HaHaWTH:
## Changes

1. Removing all username checks is very unstable and dangerous, marked as experimental and left a warning in the comment.
2. Fix Purpur's username validation config, now can actually be configured via purpur.yml
3. Add a config to enforce vanilla validations for player skulls, as skull owners with invalid characters in name will disconnect the client.


Dreeam-qwq (approved):
crab crab U

PR #558, opened 2025-11-26
Optimize Redstone Wire update logic

OPmasterLEO:
Will continue on this next week, currently getting this error:
Patch failed at 0303 Optimize Redstone Wire
error: corrupt patch at line 39
error: could not build fake ancestor


Toffikk:
what is the point of this if this is a simple copy and paste patch and also you can just enable alternative current redstone which will still be faster than this if im right

OPmasterLEO:
Its not ready yet it should be more optimized than alternative current, when done if it works out right it would be faster

MC-XiaoHei:
I'm curious where this patch came from?
If you wrote it yourself, the "build fake ancestor" issue shouldn't occur.
If you ported this from elsewhere, could you please provide a citation?

s-yh-china:
I apologize for interrupting the normal discussion, but you are not a member of LeavesMC, and the patch does not exist in any LeavesMC project.
**Please modify the package name, email, and author name you are using**, otherwise we will have no choice but to file a complaint with Github.
Is a message sent by the owner of LeavesMC.

PR #564, opened 2025-12-20
Rewrite despawn time

HaHaWTH:
Brings the ability to despawn weak-loaded entities once they are ticked, a solution for Paper#12986.


Dreeam-qwq (approved):
tql

PR #566, opened 2025-12-22
Skip saving entities if they should despawn

HaHaWTH:
Call syncEntityAge() even if despawn time is disabled, as this may be enabled in the future by users.


Dreeam-qwq (approved):
斜斜小多～

PR #570, opened 2025-12-26
Fix async mob spawn counting race in 0244 patch

Retrophy:
Switch async mob counting back to the iterator‑safe entity list and ensure iteration is properly finalized to avoid undercounting. Pass a stable iterable into createState1 and keep per‑player caps consistent so mobcaps don’t slowly drift. This keeps async spawning while restoring accurate counts and avoids thread‑unsafe access to getAllEntities.


hayanesuru:
`ServerLevel#entityTickList` excludes tracked (non-ticking but accessible) entities and is not thread-safe

`ServerLevel#getAllEntities` copies all accessible entities and is safe to move to another thread

PR #572, opened 2025-12-29
Fix configurable username check bugs

HaHaWTH:
1. Fix errors when delegating isValidPlayerName call to isReasonablePlayerName due to purpur config initialization state.
2. Use our own username regex config, Purpur's config is initialized too late.
3. Fix extra username validation check when enabling `allowOldPlayerJoin`.
4. Deprecate `removeAllCheck` configuration, use regex instead.


Dreeam-qwq (approved):
飞起来！

PR #573, opened 2026-01-01
Optimize Vec3i hashing

HaHaWTH:
Replaces the default hashCode calculation with a spatial hasher.
This implementation trades a marginal amount of calculation speed for extremely low hash collision rates, thereby improving lookup performance and mitigating worst-case hash distribution scenarios.


Dreeam-qwq (approved):
飞起来！

PR #576, opened 2026-01-03
Fix MC-17876

HaHaWTH:
This pull request addresses [MC-17876](https://bugs.mojang.com/browse/MC/issues/MC-17876), which causes attribute modifiers of equipments are applied after the entity health is set, leading to health loss on reload/rejoin.

This PR fixed the issue by loading equipment attributes before the max health is determined to prevent the health from being clamped to base value.

### To reproduce

Type command `/item replace entity @s armor.head with iron_helmet[minecraft:attribute_modifiers=[ {id:"test",type:"max_health",amount:6,operation:"add_value"}]]`, quit the server after health regen, the extra health will lost upon rejoin without this fix.


HaHaWTH:
need to be reimplemented, now drafted

PR #579, opened 2026-01-12
async-playerdata-save issues

yunuservices:
This PR fixes about data corruption ***may caused*** by async-playerdata-save feature. I didnt tested because i dont have a proper environment to do.


yunuservices:
patch tested on a environment, i didnt saw any issues with this.

PR #584, opened 2026-01-16
Apply `fix-purpur-attribute-base-patch`

VeVeVeVel:
Apply `fix-purpur-attribute-base-patch`.


Dreeam-qwq:
Hey, thanks for your contribution!
Can you provide more detail or explanation about why you are trying to apply this patch?

As far I know, this patch is to fix issues related to Model Engine mobs on Leaf, the mob's hitbox can not be overwritten correctly.

But I don't think this patch is a correct solution, that's why I moved it to the work folder, pending to check. If it is a Purpur issue (As the patch name shows), it's better to test on Purpur and report to them first. If it is a Leaf issue, better to find the real cause first.

PR #588, opened 2026-01-17
Cache world border

HaHaWTH:
The world border is only initialized once when the level is created. Lookup from data storage map multiple time is quite unnecessary and expensive.


Dreeam-qwq (approved):
小颗粒

PR #600, opened 2026-01-20
Migrate nullness annotations to JSpecify

Dreeam-qwq:
Mojang, and PaperMC have migrated nullness annotations to JSpecify. So this PR migrated JetBrains, Checker Framework's nullness annotations in Leaf patches to JSpecify.

Annotations in some upstream patches (patches from other projects, Purpur, Lithium) remain the same, to keep consistency with the upstream for better diffing on tracing updates. 

Needs to review carefully, especially for those custom collections under `org.dreeam.leaf.util` package. I'm not sure I'm doing all correct (

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/util/BlockAreaUtils.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/util/BlockAreaUtils.java
@@ -2,6 +2,7 @@ package org.dreeam.leaf.util;
 
 import net.minecraft.core.BlockPos;
 
+@org.jspecify.annotations.NullMarked
 public final class BlockAreaUtils {
     private BlockAreaUtils() {}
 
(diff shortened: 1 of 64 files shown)


hayanesuru (approved):

leaf-server/src/main/java/org/dreeam/leaf/util/BlockAreaUtils.java
```diff
@@ -2,6 +2,7 @@
 
 import net.minecraft.core.BlockPos;
 
+@org.jspecify.annotations.NullMarked
```
Dreeam-qwq: Can we move `NullMarked` to `package-info.java`? same for classes under `org.dreeam.leaf.world`.
So we don't need to add `NullMarked` to every class.

PR #603, opened 2026-01-21
configurable ice and snow tick chance

hayanesuru:


leaf-server/src/main/java/org/dreeam/leaf/world/RandomTickSystem.java
```diff
@@ -35,17 +36,19 @@ public final class RandomTickSystem {
...
+        final int randomTickSpeed = world.getGameRules().get(GameRules.RANDOM_TICK_SPEED);
         if (randomTickSpeed <= 0) {
             return;
         }
+
+        final BitRandomSource random = world.simpleRandom;
         if (!disableIceAndSnow) {
-            iceSnow(world, size, randomTickSpeed, random, raw);
+            final int rate = IceAndSnowChance.iceAndSnowChance;
+            iceSnow(world, size, randomTickSpeed, Math.max(1, rate * 2), random, raw);
```
Dreeam-qwq: Do you know why we need to `*2` here?
And I think we can remove the `Math.max`, because the rate is already `>0` during the config load phase.

Dreeam-qwq (approved):

PR #605, opened 2026-01-22
JUring implementation for future

yunuservices:
An example JUring implementation for future versions of Minecraft. This PR prepared for 26.1 version. Currently worked on 1.21.11 for testing the performance of it. `-Dleaf.enable-file-io-uring=true` for enabling the feature and `--enable-native-access=ALL-UNNAMED` for native linking. **Requires Linux environment and JDK 25**, you cant boot the server with lower than 25. When you activated the feature you should see "**JUring File I/O Enabled.**", if you activated and got failure, you will get "**JUring file I/O disabled: (error)**", also in io errors: "**JUring file I/O disabled for (path): (error)**"


yunuservices:
Note that if you want to use this in pterodactyl, you need to modify the seccomp files for enabling io_uring. Other side, in normal devices, the configuration of io_uring is easy. Also, dont forget this is an example implementation. io_uring may create attack surfaces so thats why be careful.

yunuservices:
Note that if you want to use this in pterodactyl, you need to modify the seccomp files for enabling io_uring. Other side, in normal devices, the configuration of io_uring is easy. Also, dont forget this is an example implementation. io_uring may create attack surfaces so thats why be careful.

PR #608, opened 2026-01-23
Fix LithiumHashPalette

HaHaWTH:
FAWE detects vanilla Palettes with `instanceof LinearPalette` and `HashMapPalette`, LithiumHashPalette will be considered as GlobalPalette and fall to the else branch, causing incompatibilities.


Dreeam-qwq (approved):
太强了

PR #609, opened 2026-01-26
async chunk sender: fix tracking order

hayanesuru:
add a queue to receive encoded chunk data results
copy heightmaps data on main thread
remove AsyncChunkSendThread

fix removal (change dimension)
fix entity tracking chunk packet order


Dreeam-qwq (approved):
Gooood~

PR #610, opened 2026-01-26
optimize-movement-vector-normalization

VeVeVeVel:
### **Summary**

This PR optimizes the movement vector calculation by avoiding redundant `Math.min()` and `normalize()` operations when the squared length of the vector is already within known bounds.

### **Problem**

The previous implementation always calculated the scalar length and normalized the vector, even if the vector's length was already less than or equal to the maximum threshold ($8.0$). Since `vec3.length()` involves a square root operation ($\sqrt{x}$), performing it unnecessarily can impact performance in high-frequency movement logic.

### **Changes**

I have replaced the unconditional normalization with a conditional check based on the squared length ():

* **If  ($d > 64.0$ ($8^2$)):** The vector exceeds the maximum distance. We normalize and scale it to .
* **If $d \leq 64.0$:** The vector is already within the  limit. We can use the original `vec3` directly, skipping the `length()`, `normalize()`, and `scale()` calls entirely.

### **Impact**

* **Performance:** Reduces CPU cycles by skipping square root and division operations for movement vectors with a magnitude .
* **Logic:** Maintains identical behavior to the original code but with better execution efficiency.

### DIFF
--- a/leaf-server/minecraft-patches/features/0294-optimize-movement-vector-normalization.patch
+++ b/leaf-server/minecraft-patches/features/0294-optimize-movement-vector-normalization.patch
@@ -5,21 +5,20 @@ Subject: [PATCH] optimize movement vector normalization
 
 
 diff --git a/net/minecraft/world/entity/Entity.java b/net/minecraft/world/entity/Entity.java
-index 8d092716cdcc48b829a1c0ee2e5416d648143a37..310fcb589f515b223d2b263ee99f734cd876f926 100644
+index 8d092716cdcc48b829a1c0ee2e5416d648143a37..11b9dd30605926e6cc1c1a750be945c0f3063edc 100644
 --- a/net/minecraft/world/entity/Entity.java
 +++ b/net/minecraft/world/entity/Entity.java
-@@ -1262,8 +1262,16 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
+@@ -1262,8 +1262,15 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
              double d = vec3.lengthSqr();
              if (d > 1.0E-7 || movement.lengthSqr() - d < 1.0E-7) {
                  if (this.fallDistance != 0.0 && d >= 1.0) {
 -                    double min = Math.min(vec3.length(), 8.0);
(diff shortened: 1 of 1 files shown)


leaf-server/minecraft-patches/features/0294-optimize-movement-vector-normalization.patch
```diff
@@ -0,0 +1,29 @@
...
++++ b/net/minecraft/world/entity/Entity.java
+@@ -1262,8 +1262,16 @@ public abstract class Entity implements SyncedDataHolder, DebugValueSource, Name
+             double d = vec3.lengthSqr();
+             if (d > 1.0E-7 || movement.lengthSqr() - d < 1.0E-7) {
+                 if (this.fallDistance != 0.0 && d >= 1.0) {
+-                    double min = Math.min(vec3.length(), 8.0);
+-                    Vec3 vec31 = this.position().add(vec3.normalize().scale(min));
++                    // Leaf start - optimize movement vector normalization
++                    //double min = Math.min(vec3.length(), 8.0);
++                    //Vec3 vec31 = this.position().add(vec3.normalize().scale(min));
```
Dreeam-qwq: The concept looks great. However It's better to directly replace the original code instead of commenting them, the patch already helps keep the original code part.

And I think we can inline `vec3`'s `normalize()` and `scale(8.0)` call below, to avoid unnecessary new Vec3 instance creation and useless calls.
Make sure also open a pull request to upload a benchmark to [SunBox](https://github.com/Winds-Studio/SunBox), since it's the math-related micro optimization.

Dreeam-qwq (changes requested):

VeVeVeVel:
@Dreeam-qwq I finished it.
Small Case: 40% faster
Large Case: 20% faster
```
Benchmark                            Mode  Cnt  Score   Error  Units
EntityFallClipVector.largeOptimized  avgt   10  6.550 ± 0.037  ns/op
EntityFallClipVector.largeVanilla    avgt   10  8.263 ± 0.048  ns/op
EntityFallClipVector.smallOptimized  avgt   10  4.946 ± 0.029  ns/op
EntityFallClipVector.smallVanilla    avgt   10  8.267 ± 0.178  ns/op
```

Dreeam-qwq (approved):

PR #618, opened 2026-01-31
parallel world ticking: remove buffered read

hayanesuru:
needless

#471


Dreeam-qwq:
We can't simply remove it without providing alternatives or the better solution to improve PWT's plugin compatibility.

PR #619, opened 2026-02-02
allow non-ascii character in unquoted string of command argument

hayanesuru:
test command:
```mcfunction
data merge storage ns:test {String:テスト}
data get storage ns:test String
```


Dreeam-qwq (approved):
Thanks~ I will add a config option to only enable it if `remove-all-check` is enabled.

PR #519, opened 2025-10-02
refactor async pathfinding

hayanesuru:
note: This version is unusable. `NodeEvaluator#getStart` does not support concurrency

PR #620, opened 2026-02-02
async pathfinding: fix path race condition

### LINKED ISSUE #604: Entity threw exception (ConcurrentModificationException). Async pathfinding bug?
### Checklist

- [x] I have searched the existing issues and the documentation
- [ ] I ticked all the boxes without actually reading them
- [x] I am using latest Leaf build from https://www.leafmc.one/download
- [x] I tested and couldn't replicate this issue on latest [Paper](https://papermc.io/downloads/paper) or [Purpur](https://purpurmc.org/download/purpur) build

### What behaviour is observed?

errors in console

### What is the expected behavior?

no errors

### How to reproduce?

Check https://mclo.gs/YIZ4w9c

[leaf-global.yml.txt](https://github.com/user-attachments/files/24782214/leaf-global.yml.txt)
[paper-global.yml.txt](https://github.com/user-attachments/files/24782215/paper-global.yml.txt)

### Spark profiler link

Unable to use spark without VPN (time out)

### Error log (if applicable)

https://mclo.gs/YIZ4w9c

hayanesuru:
#519 #589

Closed #604

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
@@ -1,60 +1,38 @@
 package org.dreeam.leaf.async.path;
 
-import ca.spottedleaf.moonrise.common.util.TickThread;
 import net.minecraft.core.BlockPos;
-import net.minecraft.server.MinecraftServer;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.level.pathfinder.Node;
 import net.minecraft.world.level.pathfinder.Path;
 import net.minecraft.world.phys.Vec3;
 import org.jspecify.annotations.Nullable;
 
+import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
-import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 
 /**
  * I'll be using this to represent a path that not be processed yet!
  */
-public class AsyncPath extends Path {
+public final class AsyncPath extends Path {
 
-    /**
-     * Instead of three states, only one is actually required
-     * This will update when any thread is done with the path
-     */
-    private volatile boolean ready = false;
+    private boolean ready = false;
 
-    /**
-     * Runnable waiting for this to be processed
-     * ConcurrentLinkedQueue is thread-safe, non-blocking and non-synchronized
-     */
-    private final ConcurrentLinkedQueue<Runnable> postProcessing = new ConcurrentLinkedQueue<>();
+    private final ArrayList<Consumer<Path>> postProcessing = new ArrayList<>();
 
     /**
      * A list of positions that this path could path towards
      */
     private final Set<BlockPos> positions;
 
-    /**
-     * The supplier of the real processed path
-     */
-    private final Supplier<Path> pathSupplier;
+    private @Nullable Supplier<Path> pathSupplier;
 
-    /*
-     * Processed values
-     */
+    /// Represents an asynchronous task. `null` indicates that is not ready.
+    private volatile @Nullable Path ret;
 
-    /**
-     * This is a reference to the nodes list in the parent `Path` object
-     */
-    private final List<Node> nodes;
-    /**
-     * The block we're trying to path to
-     * <p>
-     * While processing, we have no idea where this is so consumers of `Path` should check that the path is processed before checking the target block
-     */
-    private BlockPos target;
     /**
      * How far we are to the target
      * <p>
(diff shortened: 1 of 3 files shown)


leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -1,54 +1,36 @@
...
-     */
-    private volatile boolean ready = false;
+    private boolean ready = false;
 
-    /**
-     * Runnable waiting for this to be processed
-     * ConcurrentLinkedQueue is thread-safe, non-blocking and non-synchronized
-     */
-    private final ConcurrentLinkedQueue<Runnable> postProcessing = new ConcurrentLinkedQueue<>();
+    private final ArrayList<Consumer<Path>> postProcessing = new ArrayList<>();
```
MartijnMuijsers, line 24: Good change!

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -1,54 +1,36 @@
...
 
-    /**
-     * The supplier of the real processed path
-     */
-    private final Supplier<Path> pathSupplier;
-
-    /*
-     * Processed values
-     */
+    private @Nullable Supplier<Path> task;
```
MartijnMuijsers: Good change! Helps reclaim memory faster

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -176,6 +143,13 @@ public boolean canReach() {
 
     @Override
     public boolean isDone() {
+        boolean ready = this.ready;
+        if (!ready) {
+            Path ret = this.ret;
+            if (ret != null) {
+                complete(ret);
```
MartijnMuijsers: If two threads call `isDone` at the same time, they will both run `complete`. That's bad

MartijnMuijsers, line 158: If this path is only completed upon `isDone`, what happens upon `isProcessed`? Shoud it also call `complete`?

I assume `isDone()` and `isProcessed()` can only be run from the main thread, right?

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -105,47 +83,36 @@ public final void schedulePostProcessing(Runnable runnable) {
...
-            return false;
-        }
-
-        // For single position (common case), do direct comparison
-        if (positions.size() == 1) { // Both have the same size at this point
-            return this.positions.iterator().next().equals(positions.iterator().next());
-        }
-
-        return this.positions.containsAll(positions);
+        return this.positions.equals(positions);
```
MartijnMuijsers, line 94: Original is `containsAll`, this changes actual behavior

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
MartijnMuijsers: I'm assuming `schedulePostProcessing` is only called from the main thread, correct? Otherwise the `.add` will cause issues when called from multiple threads

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -87,14 +68,11 @@ public boolean isProcessed() {
...
      * Returns the future representing the processing state of this path
      */
-    public final void schedulePostProcessing(Runnable runnable) {
+    public final void schedulePostProcessing(Consumer<Path> runnable) {
         if (this.ready) {
-            runnable.run();
+            runnable.accept(this);
         } else {
-            this.postProcessing.offer(runnable);
-            if (this.ready) {
```
MartijnMuijsers, line 95: This is actually necessary, otherwise:

Thread 1 calls `schedulePostProcessing` and checks `this.ready`, it is false
Thread 2 calls `complete` and does:
```
this.ready = true;
for (Consumer<Path> consumer : this.postProcessing) {
    consumer.accept(this);
}
this.postProcessing.clear();
```
Thread 1 then does `this.postProcessing.add(runnable)` and the runnable is never executed.

MartijnMuijsers (changes requested):
.

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPathProcessor.java
```diff
@@ -67,7 +67,7 @@ protected static CompletableFuture<Void> queue(AsyncPath path) {
      */
     public static void awaitProcessing(@Nullable Path path, Consumer<@Nullable Path> afterProcessing) {
```
MartijnMuijsers, line 68: This already existed in the original, but the name `await` here is not really correct. `applyAfterProcessing`?

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -72,11 +54,10 @@ public class AsyncPath extends Path {
     public AsyncPath(List<Node> emptyNodeList, Set<BlockPos> positions, Supplier<Path> pathSupplier) {
         super(emptyNodeList, null, false);
 
-        this.nodes = emptyNodeList;
         this.positions = positions;
-        this.pathSupplier = pathSupplier;
+        this.task = pathSupplier;
 
-        AsyncPathProcessor.queue(this);
+        AsyncPathProcessor.queue(() -> this.ret = pathSupplier.get());
```
MartijnMuijsers: Can we change this to:

```java
() -> {
    if (this.ret == null) {
        this.ret = pathSupplier.get();
    }
}
```

The `this.ret == null` is only a heuristic that may lead us to skip this if the main thread did it in `process` before the `AsyncPathProcessor` got to it

leaf-server/src/main/java/org/dreeam/leaf/async/path/AsyncPath.java
```diff
@@ -105,47 +83,36 @@ public final void schedulePostProcessing(Runnable runnable) {
...
-        Runnable runnable;
-        while ((runnable = this.postProcessing.poll()) != null) {
-            if (isTickThread) {
-                runnable.run();
-            } else {
-                MinecraftServer.getServer().scheduleOnMain(runnable);
-            }
+        final Path ret = this.ret;
+        final Supplier<Path> task = this.task;
+        final Path bestPath = ret != null ? ret : Objects.requireNonNull(task).get();
```
MartijnMuijsers:
```suggestion
        final Path bestPath = ret != null ? ret : (this.ret = Objects.requireNonNull(task).get());
```

So that we can use the heuristic I mentioned above ^

MartijnMuijsers (approved):

Dreeam-qwq (approved):
Nice PR! Thanks~

PR #624, opened 2026-02-02
optimize `FlowingFluid#getHeight`

hayanesuru:


VeVeVeVel (approved):
I'm using this PR and haven't had any issues. I haven't run any benchmarks, but it looks good theoretically.

Dreeam-qwq:
The condition `state.getOwnHeight() == 1.0F` will never be true, the maximum height property of flowing/source lava or water is 8, instead of 9. Unless plugins inject custom fluid height, but that's the very rare case.

PR #632, opened 2026-02-07
Fixed skipping inactive entity condition

wling-art:
Like the title said


HaHaWTH (changes requested):
The paper config `world.paperConfig().entities.markers.tick` should be checked as well (for short circuit condition), always retrieve markers if marker ticking is set to false.

HaHaWTH (approved):

PR #633, opened 2026-02-08
Avoid sync chunk load in dismountVehicle method

BoocordStudios:
Modified LivingEntity#dismountVehicle to avoid forcing a sync chunk load when dismounting from a vehicle. This change uses getBlockStateIfLoaded to check the vehicle's chunk state, preventing potential deadlocks in Parallel World Ticking.

Logs of the issue: https://mclo.gs/wHZApeJ


Dreeam-qwq:
Needs more info to figure out why this thread deadlock can happen.

PR #636, opened 2026-02-09
Fix MC-17876 on 1.21.11

HaHaWTH:
Previous PR #576.

Fixed by temporarily bypassing the max health check during entity creation, less invasive than previous one.

### To reproduce

Type command `/item replace entity @s armor.head with iron_helmet[minecraft:attribute_modifiers=[ {id:"test",type:"max_health",amount:6,operation:"add_value"}]], quit the server after health regen, the extra health will lost upon rejoin without this fix`.


Dreeam-qwq (approved):
Jump jump

PR #637, opened 2026-02-09
Further optimize getBiome

HaHaWTH:
Eliminating fp divisions brings about 45% improvement in retrieving QuartPos.

Benchmark results (batching 4096 operations per invocation):

```
Benchmark                              Mode  Cnt     Score    Error  Units
BiomeLookupBenchmark.testFpDivision  avgt    5  2746.521 ± 23.079  ns/op
BiomeLookupBenchmark.testTableLookup    avgt    5  1549.182 ± 28.602  ns/op
```

### DIFF
--- a/leaf-server/minecraft-patches/features/0097-Carpet-Fixes-Optimized-getBiome-method.patch
+++ b/leaf-server/minecraft-patches/features/0097-Carpet-Fixes-Optimized-getBiome-method.patch
@@ -9,8 +9,10 @@ Original project: https://github.com/fxmorin/carpet-fixes
 Optimized the getBiome call to be 25% - 75% faster
 This is a fully vanilla optimization.
 
+Co-authored-by: HaHaWTH <102713261+HaHaWTH@users.noreply.github.com>
+
 diff --git a/net/minecraft/world/level/biome/BiomeManager.java b/net/minecraft/world/level/biome/BiomeManager.java
-index 73962e79a0f3d892e3155443a1b84508b0f4042e..a48175a7ebb1788ace46395621ed78d910178a53 100644
+index 73962e79a0f3d892e3155443a1b84508b0f4042e..93cbe1d76efd97899ea520a92e88b94639824174 100644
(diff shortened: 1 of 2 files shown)


hayanesuru (approved):

leaf-server/minecraft-patches/features/0097-Carpet-Fixes-Optimized-getBiome-method.patch
```diff
@@ -21,9 +23,11 @@ index 73962e79a0f3d892e3155443a1b84508b0f4042e..a48175a7ebb1788ace46395621ed78d9
  
      public BiomeManager(BiomeManager.NoiseBiomeSource noiseBiomeSource, long biomeZoomSeed) {
          this.noiseBiomeSource = noiseBiomeSource;
-@@ -29,39 +30,67 @@ public class BiomeManager {
+@@ -28,40 +29,73 @@ public class BiomeManager {
+         return new BiomeManager(newSource, this.biomeZoomSeed);
      }
  
++    private static final double[] QUART_OFFSETS = new double[]{0.0, 0.25, 0.5, 0.75}; // Leaf - Optimized getBiome method - eliminate divisions
```
OverwriteMC, line 30: Can't it be moved to me.titaniumtown.ArrayConstants?

HaHaWTH: I'd prefer not to. ArrayConstants is for reducing array allocations by reusing empty arrays. This table is specific to the math logic of this method, so it stays near the code that uses it.

OverwriteMC: > I'd prefer not to. ArrayConstants is for reducing array allocations by reusing empty arrays. This table is specific to the math logic of this method, so it stays near the code that uses it.

yeah, I thought about it, and it's would not be entirely appropriate

PR #647, opened 2026-02-18
Reduce AbstractContainerMenu allocations

HaHaWTH:
Allocating Supplier in `broadcastChanges` every loop causes high GC pressure and also hurts performance with a high player count:

image


Dreeam-qwq (approved):
tql

PR #659, opened 2026-03-01
async pathfinding: fix brain process path

hayanesuru:
move process to tick
inline AcquirePoi and SetClosestHomeAsWalkTarget


Dreeam-qwq (approved):

Dreeam-qwq:
Thanks~ And I think the inline part can move to another patch later?

PR #665, opened 2026-03-09
Fix cross-dimension entity travel under parallel world ticking with post-tick handoff

splatage:
### Summary
This PR fixes cross-dimension entity portal travel when parallel world ticking (PWT) is enabled.

The core issue was that entities could begin portal travel while ticking on one world thread, then immediately read or mutate destination-world state from the wrong thread. That caused hard thread-check failures and made cross-dimension travel unsafe under PWT.

This patch changes that flow so cross-dimension transfers are queued during world tick and then drained after all world tick tasks join, on the safe server-thread handoff phase. This keeps destination-world mutation out of source-world tick threads and preserves normal same-dimension behaviour.

### Problem
Under PWT, portal handling could attempt cross-world work inline during entity ticking. That caused problems such as:
- destination-world POI / portal lookup from the wrong world thread
- cross-dimension teleport mutation happening during source-world tick
- unsafe behaviour when entities changed dimension while still inside the parallel tick phase

This was reproducible with entities entering Nether portals while PWT was enabled.

### What this PR changes
**Entity portal handling**
- Detects Nether / End portal travel during entity portal processing
- Prevents inline cross-dimension completion during parallel world ticking
- Queues a pending dimension transfer instead

**Server-side transfer handoff**
- Adds a pending transfer queue on the server
- Drains queued transfers after world tick futures complete
- Performs destination lookup and final teleport from the safe handoff phase

**Transfer validation / safety**

Before applying a queued transfer, the drain phase verifies that:
- the entity still exists
- the entity is still in the expected source world
- the portal process still matches the queued transfer
- plugin cancellation state is respected
- normal `isAllowedToEnterPortal(...) / canTeleport(...)` checks still pass

**Passenger handling**

The latest patch in this PR also fixes entities with passengers so they no longer disappear during portal travel. Passenger-bearing travel now behaves correctly through the queued cross-dimension handoff path.

**Scope**

This PR is intentionally limited to:
- portal-driven cross-dimension travel
- PWT-safe handoff for those transfers
- passenger-bearing entity travel fixes

It does not attempt to broadly solve every cross-world command or side-effect case under PWT. That remains separate work.


splatage:
Apologies, some of the file edits were not included in the patch set.

PR #675, opened 2026-03-18
[Optimize] Add options to reduce unnecessary packets

VeVeVeVel:
## Description
This PR applies an optimization patch from the [Canvas](https://github.com/CraftCanvasMC/Canvas) project to Leaf. It introduces a rough filter for `ClientboundSetEntityMotionPacket`, which can reduce bandwidth usage by up to 60%.
https://github.com/CraftCanvasMC/Canvas/blob/ver/1.21.11/canvas-server/minecraft-patches/sources/net/minecraft/server/level/ServerEntity.java.patch

While entirely disabling this packet causes major desync issues, this patch carefully filters the packets while retaining them for specific entities that rely on them for smooth visual gameplay.

### Implementation Details
The packet is filtered out by default when the `ReduceUselessPackets.filterClientboundSetEntityMotionPacket` config is enabled, **except** for the following entities:

1. **Squids**: Required as they are mobs with a 360° range of rotation.
2. **Item Entities**: Disabling this causes glitches and client-side desyncs (often within 0.5 blocks). Packets are only sent if the item has actually moved (`flag1`).
3. **Ender Eyes**: The client does not have access to the Stronghold structure data to calculate velocity on its own, so it relies on the server sending the velocity via this packet.
4. **Shulker Bullets**: Since their trajectory uses randomness, we must send velocity packets to ensure their movement remains smooth on the client.

## Credits
* **Original Project:** [CraftCanvasMC/Canvas](https://github.com/CraftCanvasMC/Canvas)
* **Original Author:** Dueris
* **Original License:** GPL v3


leaf-server/minecraft-patches/features/0306-Filter-ClientboundSetEntityMotionPacket.patch
```diff
@@ -0,0 +1,73 @@
...
+                         }
++                        // Canvas end - filter ClientboundSetEntityMotionPacket
+                     }
+                     } // SparklyPaper end
+                 }
+@@ -555,9 +566,20 @@ public class ServerEntity {
+                                     new ClientboundSetEntityMotionPacket(this.entity.getId(), this.lastSentMovement),
+                                     new ClientboundProjectilePowerPacket(abstractHurtingProjectile.getId(), abstractHurtingProjectile.accelerationPower)
+                                 )));
+-                            } else {
```
Dreeam-qwq, line 48:
```suggestion
-                            } else if (!org.dreeam.leaf.config.modules.opt.ReduceUselessPackets.filterClientboundSetEntityMotionPacket || (conditions)) {
```
You can merge these to single-line condition like this.

VeVeVeVel: Thanks for the suggestion! I've merged it into a single-line condition as you recommended. I've also added the documentation for this configuration.

VeVeVeVel:
## Description
This PR adds the `disable-useless-particles` configuration to reduce unnecessary network traffic and server load. 

**About Disable Useless Particles:**
This feature disables the server-side logic for the following cosmetic particles:
* Sprint Particles
* Death Particles
* Effect Particles
* Water Splash Particles
* Bubble Column Particles

As noted by the original developer of this patch:
> "It is recommended to enable these options, as the client displays most of these particles already, so the server-side particle logic is not needed."
https://github.com/CraftCanvasMC/Canvas/commit/f43bf7696dde9d04fde7d5ec118cb481ce4a686e
https://github.com/CraftCanvasMC/Canvas/blob/ver/1.21.11/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
https://github.com/CraftCanvasMC/Canvas/blob/ver/1.21.11/canvas-server/minecraft-patches/sources/net/minecraft/world/entity/LivingEntity.java.patch

By disabling this unneeded server-side logic, we can effectively save network bandwidth, all while maintaining the exact same visual experience for players.

## Credits
* **Original Project:** [CraftCanvasMC/Canvas](https://github.com/CraftCanvasMC/Canvas)
* **Original Author:** Dueris
* **Original License:** GPL v3

Dreeam-qwq (approved):

hayanesuru (approved):

PR #687, opened 2026-03-26
 Do not process VehicleEntityCollisionEvent if no listeners

OverwriteMC:
Lag machines using many minecarts in one place?
Please


OverwriteMC:
Nah

PR #688, opened 2026-03-27
Do not process VehicleEntityCollisionEvent if no listeners

OverwriteMC:
correct patch


Dreeam-qwq (changes requested):
"optimize empty check for passengers" changes better to move to `Optimize-map-lookups-with-isEmpty-check.patch`, be separated with this patch.

Dreeam-qwq (approved):

PR #693, opened 2026-04-01
Optimize empty passengers check

OverwriteMC:
not the cleanest, but the most effective approach


Dreeam-qwq (approved):
Looks goood. But next time, remember to create a new branch instead of using the main branch.

PR #701, opened 2026-04-10
optimize CompoundTag map compares

hayanesuru:
#646

avoid map entry and iterator overhead

inline `StringCanonizingOpenHashMap#deepCopy`

### DIFF
--- a/leaf-server/minecraft-patches/features/0171-Further-reduce-memory-footprint-of-CompoundTag.patch
+++ b/leaf-server/minecraft-patches/features/0171-Further-reduce-memory-footprint-of-CompoundTag.patch
@@ -7,7 +7,7 @@ Original license: GPL-3.0-only
 Original project: https://github.com/embeddedt/ModernFix
 
 diff --git a/net/minecraft/nbt/CompoundTag.java b/net/minecraft/nbt/CompoundTag.java
-index c9c050966924ddd315e7829e7f3aee2afacc106b..6e859b1add79ce4e636e2994f59963da2cb440a9 100644
+index c9c050966924ddd315e7829e7f3aee2afacc106b..1971fcfbc1c79ab3ebba612f1999e5a206600ebe 100644
 --- a/net/minecraft/nbt/CompoundTag.java
 +++ b/net/minecraft/nbt/CompoundTag.java
 @@ -54,7 +54,7 @@ public final class CompoundTag implements Tag {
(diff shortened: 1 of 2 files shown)


leaf-server/minecraft-patches/features/0171-Further-reduce-memory-footprint-of-CompoundTag.patch
```diff
@@ -28,13 +28,19 @@ index c9c050966924ddd315e7829e7f3aee2afacc106b..6e859b1add79ce4e636e2994f59963da
...
      @Override
-@@ -400,6 +400,11 @@ public final class CompoundTag implements Tag {
+@@ -400,6 +400,17 @@ public final class CompoundTag implements Tag {
  
      @Override
      public CompoundTag copy() {
 +        // Leaf start - Further reduce memory footprint of CompoundTag
 +        if (this.tags instanceof org.dreeam.leaf.util.map.StringCanonizingOpenHashMap<Tag> stringCanonizingTags) {
-+            return new CompoundTag(org.dreeam.leaf.util.map.StringCanonizingOpenHashMap.deepCopy(stringCanonizingTags, Tag::copy));
++            org.dreeam.leaf.util.map.StringCanonizingOpenHashMap<Tag> ret = new org.dreeam.leaf.util.map.StringCanonizingOpenHashMap<>(stringCanonizingTags.size(), 0.8f);
```
HaHaWTH, line 53: I'm curious about how much improvement this patch can bring, as we ignored the load factor of the original map when copying, rehashing cost is expensive.

I suggest add another method `getLoadFactor` in the `StringCanonizingOpenHashMap` to reuse load factor

leaf-server/src/main/java/org/dreeam/leaf/util/map/StringCanonizingOpenHashMap.java
```diff
@@ -42,19 +40,29 @@ public void putAll(Map<? extends String, ? extends T> m) {
...
-            Map.Entry<String, T> entry = iterator.next();
-            newMap.putWithoutInterning(entry.getKey(), deepCopier.apply(entry.getValue()));
+    @Override
+    public boolean equals(Object o) {
+        if (o == this) return true;
+        if (!(o instanceof Map)) return false;
+        final Map<?, ?> m = (Map<?, ?>)o;
+        if (m.size() != size()) return false;
+        if (containsNullKey) {
+            if (!value[n].equals(m.get(key[n]))) {
```
HaHaWTH, line 35: Should we use `Objects.equals` here to guard against null values? And use `m.containsKey(key)` to ensure the key exists, `null == null` is always true

HaHaWTH (changes requested):

Dreeam-qwq:
I think we can drop this patch directly, see
https://github.com/Winds-Studio/Leaf/pull/720
https://github.com/embeddedt/ModernFix/commit/71b444288bcff33fa74bef0a182f30d1cc500f33

However, thanks for your PR!

PR #702, opened 2026-04-12
remove duplicated purpur code

OverwriteMC:
that is actually a bug


Dreeam-qwq:
It's better to move it to `Fix-Pufferfish-and-Purpur-patches.patch`, I'll do it for you.

Dreeam-qwq (approved):
Thanks to point it out~

PR #703, opened 2026-04-13
Make EvictingRingList support dynamic growth

HaHaWTH:
Update tableSizeFor to use latest Java features

Dynamic grow the list size to reduce initial memory usage


Dreeam-qwq (approved):
Jump!

PR #706, opened 2026-04-16
Skip item merge checks for full stacks

nostalfinals:
By default, when `ItemEntity#mergeWithNeighbours` is called, it will continue calling `tryToMerge` even after the current item entity is already full.

Although `tryToMerge` uses `areMergable` to check whether the merge would exceed the maximum stack size before actually calling `merge`, those checks are guaranteed to fail once the current item entity itself is already full.

This patch avoids those guaranteed-to-fail merge checks by exiting the neighbour scan early once the current item entity has already reached its maximum stack size.

This does not break `ItemMergeEvent`, because that event is only fired after `areMergable` has already passed, which would never happen for these guaranteed-to-fail cases.


Dreeam-qwq:
Thanks! Looks good.

But I can push a change later to prevent calling `getItem` multiple times.

Dreeam-qwq (approved):

PR #710, opened 2026-04-22
Limit dont-save entity config to chunk serialization only

XUANHLGG:
This option were meant to prevent falling blocks and primed TNT from being written to chunk entity data, but the old implementation applied them through shouldBeSaved(), which is also used during chunk unload. As a result, those entities could bypass the normal unload/remove path and remain in memory. This change limits the handling to saveEntityChunk0(), so falling blocks and primed TNT are still unloaded normally while remaining excluded from chunk saves.
#606


Dreeam-qwq:
Hi, looks like you linked a Leaf 1.21.4 issue. However, Leaf 1.21.4 is marked as unsupported and will not receive any updates. Can you replicate this issue on the latest Leaf 1.21.11?

PR #717, opened 2026-04-28
remove unnecessary getBukkitEntity calls

OverwriteMC:


Dreeam-qwq:
The `getBukkitEntity` call is cheap enough. It's a lazy init in this method, and the `bukkitEntity` will not be set to null in other places. These duplicated calls can be ignored, unless there is an actual performance issue shown in the spark profiler.

However, thanks for submitting this PR!

PR #725, opened 2026-05-02
Add Leaves's Lithium sleeping block entity

Dreeam-qwq:
Lithium's sleeping block entity is a very useful optimization for block entity ticking. It can give a decent performance improvement, especially for SMP or "生电" servers, and keep vanilla parity at the same time.

Based on the Leaves's version of Lithium Sleeping Block Entity, since their impl contains fixes to the events compatibility issues.
The latest sleeping block-entity related changes were synchronized from Lithium, which is separated as a temp patch. Once Leaves finishes the 26.x update, I will open a Pull Request to merge these changes to them.

The original hopper related optimizations (Gale's hopper suck item interval and Leaf's `throttle-hopper-when-full`) are dropped, since there is no need if Lithium's sleeping block entity optimization exists.

Additionally, Leaves's vanilla hopper is also added.

Still need to test for several days.

### Config changes

Set `sleeping-block-entity` to true to enable Lithium's sleeping block entity optimization.
Set `use-vanilla-hopper` to true to allow the use of more redstone builds, such as item collection.
```
performance:
  sleeping-block-entity: false
gameplay-mechanisms:
  use-vanilla-hopper: false
```


Dreeam-qwq:
Tested on several servers (Tested with BlockLocker and QuickShop plugins), and didn't have any issues so far.

HaHaWTH (approved):
FINALLY ARRIVED

PR #730, opened 2026-05-05
optimize `collectSpawningChunks`

hayanesuru:
reduce hash lookup on high simulation distance


Dreeam-qwq (approved):
Looks goood.
Whether need to change the name of `getMiss` for better naming? Or the current one is fine?

PR #743, opened 2026-05-17
a optimization for prac servers

nemyontop:
i am quite new so i added entity limiter by chunks , only allow plugin spawns (like mannequins for practice bot)


Dreeam-qwq:
Entity limitation by chunk this kind of features can be easily done using plugins and in a more customized method (You define limitations for each entity type).

PR #748, opened 2026-05-18
Refactor hashCompare to use try-with-resources

kayerajava:
Wrapped "FileInputStream" in a try-with-resources block to ensure the stream 
is always closed after use, even if an exception occurs.
Prevents file descriptor leaks under heavy load and fixes potential file lock issues on Windows.


Dreeam-qwq:
Hi, changes to Leaves should be made directly to Leaves instead. Once they merge it, we will sync related changes back to Leaf later. Thanks for your understanding!

PR #750, opened 2026-05-21
fix OptimizedPoweredRails thread safety crash under PWT

Nrleryxx:
Replace the shared static CHECKED_POS_POOL with per-call local Object2BooleanOpenHashMap instances.
Under Parallel World Threading (PWT), multiple threads accessing the static map
simultaneously causes ConcurrentModificationException / data races. Each method
call now creates its own local map, eliminating shared mutable state entirely.
No behavioral change to rail power logic.


Dreeam-qwq (approved):
Thanks!

PR #752, opened 2026-05-25
Reduce debug subscribers overhead

HaHaWTH:
Vanilla ServerDebugSubscribers is causing massive object allocations even when it's not in use.

1. Throttle ticking of subscribers, update every single second is enough.
2. Add system property `Leaf.disable-vanilla-debug-feature` to completely disable debug flags.


Dreeam-qwq (approved):
飞起来！

PR #763, opened 2026-06-04
Fix lag-compensation causing cobblestone generators to produce stone

Nrleryxx:
Fix lag-compensation water tick delay breaking cobblestone generators at low TPS; remove unused enable-for-water config.


Dreeam-qwq:
Thanks for your contribution!
However, simply removing this option is not a correct fix. If this is unfixable, we can note it in the Leaf docs, that this feature may break some vanilla redstone builds.

Also, you can check https://github.com/Winds-Studio/Leaf/pull/764 for a reference.

PR #767, opened 2026-06-06
Fix some ABA races of Linear region format

HaHaWTH:
Preparing to sync following changes from Luminol:

1. https://github.com/LuminolMC/Luminol/commit/3ca8eba57d252d1fac75fb0308199e48e146465a

Add read-only mode for Linear v2 for emergency use.


Dreeam-qwq (approved):
看饿了

PR #771, opened 2026-06-10
Add configurable shield blocking delay

vhs03:
Adds a new shield blocking delay option to leaf-global.yml

Shields normally take 5 ticks before they start blocking after being raised and there was no easy way to change that anymore after Paper removed their option for it in the 1.21.5 update

This adds the setting back in Leaf so you can control the delay again

Lower values make shields feel more responsive which is useful for pvp servers where even a few ms can make a difference


vhs03:
Im aware that you can already do this now by using data component
The difference is that this only affects the specific shields where the component is actually changed
It does not automatically apply to every normal shield on the server such as crafted shields
This setting is more convenient when you just want all shields to be more responsive globally without having to replace every shield or make custom items for it

This basically just changes the default value

vhs03:
This just changes the default value which is a lot easier than making a custom plugin which permanently has to detect and edit the data components of every shield

So this change makes more sense in the server software rather than in a plugin

PR #779, opened 2026-06-16
Ensure NearbyPlayers add/remove operations running on main

HaHaWTH:
Draft, don't merge.

Current status: Add thread check to locate the root cause


Dreeam-qwq:
Closed since the original issue no longer happens.

PR #784, opened 2026-06-21
Guard navigation against unloaded target chunks

Smorki:


Smorki:
This fixes a crash during async teleportation where navigiation can tick toward a target before the destination chunk is fully loaded. It adds a loaded-chunkguard before accepting the processed path.

Smorki:
Hey

Smorki:
İs anyone there?

Dreeam-qwq:
> This fixes a crash during async teleportation where navigiation can tick toward a target before the destination chunk is fully loaded. It adds a loaded-chunkguard before accepting the processed path.

Hi, can you send the stacktrace of the crash, or the full log at that time?

Btw, I assume it is handled by the chunk system, so this loaded-chunk guard becomes useless under this scenario. Did you test in production that this get chunk condition can actually be met?

Nrleryxx:
crab crab

ichbinrather:
crab crab

Smorki:
I am currently away from my development setup, so I can't pull the exact stacktrace right now, but I will send it as soon as I can.
Under heavy load, async teleportation sometimes outruns the chunk loading threads, creating a race condition where the navigation ticks before the chunk system finishes registration. This guard was meant as a safety layer for that specific edge case.
I will double-check my environment logs and share the details with you soon!

Smorki:
To clarify: Yes, I tested this under a production like load. Without this guard, I was consistently hitting a crash when async tasks attempted to access a chunk that was in the middle of a registration cycle. The vanilla chunk system wasn't preventing the access in that specific race condition window. This patch directly resolves that crash in my environment.

Dreeam-qwq:
> To clarify: Yes, I tested this under a production like load. Without this guard, I was consistently hitting a crash when async tasks attempted to access a chunk that was in the middle of a registration cycle. The vanilla chunk system wasn't preventing the access in that specific race condition window. This patch directly resolves that crash in my environment.

Again, can you send the error stacktrace or crash report? Because you said "I was consistently hitting a crash", I assume it's easy to replicate the crash in your environment?

PR #785, opened 2026-06-22
fix async pathfinding tick rate

hayanesuru:
cache paper sensor & behavior tick rate

remove no effect `CompletableFuture#orTimeout`

replace `NearestBedSensor` tick rate


Dreeam-qwq (approved):

noramibu (changes requested):
IMHO this looks risky because it can make async pathfinding run on the main server thread again

From what I understand, old logic only used pending path when the async worker had already finished it, but new logic tries to finish the pending path immediately. If the async worker has not finished yet, main thread may end up waiting for it or doing the path calculation itself and under load it may cause TPS/MSPT spikes, especially with villagers searching for beds or job sites 

I believe a safer flow would be: if the pending path is ready, use it; if it is not ready yet, keep it pending and try again later ^^

Dreeam-qwq (approved):
Looks good.
About the potential performance regression. I think it's fine, if we really got big regression in the production, we can improve it later.

PR #794, opened 2026-06-30
Fix all chunks slime chunk

noramibu:
Fixes Paper's "all-chunks-are-slime-chunks" config being ignored by Secure Seed's slime chunk check due to ternary precedence ^^

Currently, even when Secure Seed is disabled or enabled, "all-chunks-are-slime-chunks" does not work, this causes the server to use the secure slime chunk check instead of treating every chunk as a slime chunk


Dreeam-qwq (approved):
Crab crab

noramibu:
crab crab

PR #795, opened 2026-07-01
Fix skeleton bow accuracy reload

noramibu:
Fixes skeleton "mobs.skeleton.bow-accuracy" being reset after "/purpur reload".

Leaf is adding an optimization to skip Purpur JS expression evaluation when the value is default, but it was checking the previous loaded value instead of checking the real default value

Because of this, if the same custom value was loaded twice, Leaf thought it was default and stopped using the custom accuracy.

Steps to reproduce:

1. Set "mobs.skeleton.bow-accuracy" to  an obvious value, such as 0
2. Start the server
3. Spawn a skeleton and see that it shoots very accurately
4. Change "mobs.skeleton.bow-accuracy" to another obvious value such as 100
5. Run "/purpur reload" command
6. Skeleton now misses almost every shot
7. Run "/purpur reload" command again without changing the configuration we set on step 4
8. Skeleton will start shooting accurately again, like the custom value was completely ignored

This patch checks against Purpur's real default value instead, so custom bow accuracy keeps working after reload ^^


Dreeam-qwq (approved):
Nice catch! Thanks~

PR #801, opened 2026-07-05
Update 0194-SparklyPaper-Parallel-world-ticking.patch

ichbinrather:
Title: Fix explosionDensityCache concurrent access crash with PWT

Description:

When PWT (Parallel World Ticking) is enabled, every explosion (creeper, TNT, bed, respawn anchor) can crash the server or corrupt the explosionDensityCache, causing explosions to deal wrong damage or destroy wrong blocks.

Root cause: In MinecraftServer.tickChildren(), after submitting the world tick to the executor, the main thread immediately called level.explosionDensityCache.clear(). The world thread was still reading/writing the same HashMap at the same time, causing a data race.

Fix: Moved the clear() call to after all worlds finish ticking, so there is no concurrent access to the HashMap.

### DIFF
--- a/leaf-server/minecraft-patches/features/0194-SparklyPaper-Parallel-world-ticking.patch
+++ b/leaf-server/minecraft-patches/features/0194-SparklyPaper-Parallel-world-ticking.patch
@@ -235,13 +235,15 @@ index 9077a799f5113ce598d1f5c6c72bfec8cfb8f747..0bd82f3e378f7408ae029ae3b68a6273
 +                tickLevel(level, haveTime);
              }
  
-             profiler.pop();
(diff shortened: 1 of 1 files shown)


Dreeam-qwq:
The core idea is correct; thanks for pointing it out!
There is a small issue in the current PR: you just moved the cache clear to the inside of the `while (!tasks.isEmpty()) {`, but if the parallel world ticking config is disabled, the cache will never get cleared.

And I'm considering introducing the parallel world ticking implementation from [Fish](https://github.com/Biquaternions/Fish)

Not sure whether I will merge this PR, since Fish already includes this fix.

PR #805, opened 2026-07-09
async tracker: exclude packets before starting track

hayanesuru:
another fix is send packets to packet queue of async tracker


Dreeam-qwq:
Should be superseded by https://github.com/Winds-Studio/Leaf/pull/905

PR #815, opened 2026-07-15
Cache method calls in HopperHelper#determineComparatorUpdatePattern

Smorki:
motivation
hopperhelper.determinecomparatorupdatepattern calls getcontainersize and getmaxstacksize repeatedly inside loops since container is an interface this causes redundant virtual dispatch calls that jit can't always inline this overhead adds up on hopper heavy servers

changes
cached getcontainersize and getmaxstacksize into local variables at the start

updated all loop calls to use these cached values

impact
removes unnecessary interface method dispatching

maintains identical behavior with lower overhead per call

provides measurable performance gains in hopper heavy scenarios

### DIFF
--- a/leaf-server/src/main/java/net/caffeinemc/mods/lithium/common/hopper/HopperHelper.java
+++ b/leaf-server/src/main/java/net/caffeinemc/mods/lithium/common/hopper/HopperHelper.java
@@ -101,20 +101,21 @@ public class HopperHelper {
         float contentWeight = 0f;
         int numOccupiedSlots = 0;
 
-        // Leaf - cache method call results to avoid repeated interface dispatch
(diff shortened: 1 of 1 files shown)


Smorki:
hi

Smorki:
@Dreeam-qwq  @MartijnMuijsers

Dreeam-qwq (approved):
Looks good, thanks!

However, you only add one comment, but you have multiple changes. But it's fine, I can fix it during the merge.

Smorki:
Noted, thanks for reviewing 🙏 will add per change comments next time

Smorki:
btw can i have role in dc please? ;  .smorki

PR #817, opened 2026-07-16
Optimize MutableBlockPos in CombinedHeightmapUpdate loop

Smorki:
## Summary

Small optimization in `CombinedHeightmapUpdate#updateHeightmaps` to avoid redundant field writes on the `MutableBlockPos` inside the search loop.

## Changes

- Initialize `MutableBlockPos` with `x` and `z` already set (they never change during the loop).
- Use `setY(searchY)` inside the loop instead of `set(x, searchY, z)`.

## Rationale

The loop can iterate up to the full world height (`y - 1` down to `bottomY`, so potentially hundreds of iterations for tall worlds). In each iteration, only `searchY` changes — `x` and `z` are loop-invariant.

- Before: every iteration writes 3 fields (`x`, `y`, `z`) via `set(x, searchY, z)`.
- After: every iteration writes only 1 field (`y`) via `setY(searchY)`.

This removes 2 redundant field writes per iteration. While JIT may hoist some of these, expressing the intent explicitly in the code is beneficial and avoids relying on optimizer behavior.

### DIFF
--- a/leaf-server/src/main/java/net/caffeinemc/mods/lithium/common/world/chunk/heightmap/CombinedHeightmapUpdate.java
+++ b/leaf-server/src/main/java/net/caffeinemc/mods/lithium/common/world/chunk/heightmap/CombinedHeightmapUpdate.java
@@ -95,11 +95,11 @@ public final class CombinedHeightmapUpdate {
             return;
         }
 
-        BlockPos.MutableBlockPos mutable = new BlockPos.MutableBlockPos();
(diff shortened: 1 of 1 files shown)


Dreeam-qwq:
I think there's almost no difference in saving two value assignments.
Thanks for your contribution!

Smorki:
Thanks; Ill pay attention to hot path areas from now on

PR #819, opened 2026-07-16
Avoid redundant MutableBlockPos writes in ComparatorTracking

Smorki:
## Summary

Small cleanup in `ComparatorTracking` to avoid redundant `MutableBlockPos` writes.

## Changes

 Hoist pos.getX/Y/Z() and `Direction#getStepX/Z()  out of the loops.
 Replace set(pos) + move(direction, offset) with a single set(x, y, z)

Applied in both  notifyNearbyBlockEntitiesAboutNewComparator  and  findNearbyComparators .


Smorki:
last PR of the night 😴

Dreeam-qwq:
Hi, thanks for your contribution!
Checked the related code. I think `ComparatorTracking#notifyNearbyBlockEntitiesAboutNewComparator` is not very hot, no need to do this kind of micro optimization, the benefit will be very tiny.

`ComparatorTracking#notifyNearbyBlockEntitiesAboutNewComparator` method is for tracking nearby comparator blocks for the sleeping block entity feature, and will be called from the `BlockBehavior#onPlace` under `LevelChunk#setBlockState`. The comparator block can not be pushed by the piston, so this method will only be called when a player places a comparator block, so it's not a hot path.

Smorki:
Thanks; Ill pay attention to hot path areas from now on

PR #835, opened 2026-07-27
Replace DiscardPolicy with CallerRunsPolicy in AsyncPlayerDataSaving

ichbinrather:
DiscardPolicy silently drops player data save tasks when the pool is overloaded. This can cause player inventory/XP/location loss.

Replaced with CallerRunsPolicy so rejected tasks execute on the submitting thread instead of being silently discarded.


hayanesuru:
queue slot is not limited

PR #837, opened 2026-07-28
Move ClientboundCommandsPacket creation off the main thread

purpurcof:
## How do I reproduce this?

Join an server with FAWE and WorldGuard installed (or any setup with hundreds/thousands of commands). Profile the main thread with spark during login - `ClientboundCommandsPacket.<init>()` shows up with ~400ms, split across `enumerateNodes()` (BFS + Object2IntOpenHashMap) and `createEntries()`.

## Root cause

`Commands#sendCommands()` constructs `new ClientboundCommandsPacket(rootCommandNode, ...)` synchronously on the netty event loop (which is the main server thread). The constructor traverses the entire command tree via BFS, builds an Object2IntOpenHashMap, and serialises every node - O(n) in the number of commands. With FAWE + WorldGuard adding thousands of command nodes, this single constructor call blocks the main thread for ~394ms per login.

## Fix

Move packet construction to `COMMAND_SENDING_POOL` (an async executor already used elsewhere in `Commands.java`). `connection.send()` is scheduled back on the main server thread via `MinecraftServer.getServer().execute()` to keep `Connection.pendingActions` single-thread-safe.

## Verification

- Spark with patch https://spark.lucko.me/sh1RlW5XLJ

## References

- Related to https://github.com/Winds-Studio/Leaf/pull/656
- Spark without patch https://spark.lucko.me/Z2uPCUtZ3R

### DIFF
--- a/leaf-server/minecraft-patches/features/0335-Move-ClientboundCommandsPacket-creation-off-the-main.patch
+++ b/leaf-server/minecraft-patches/features/0335-Move-ClientboundCommandsPacket-creation-off-the-main.patch
@@ -4,17 +4,18 @@ Date: Mon, 27 Jul 2026 12:00:00 +0300
 Subject: [PATCH] Move ClientboundCommandsPacket creation off the main thread
 
 Build the packet on the async command sending pool to avoid
-enumerateNodes()/createEntries() lag on login with many commands.
+enumerateNodes()/createEntries() lag on login with many plugin commands.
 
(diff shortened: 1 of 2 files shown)


Dreeam-qwq (approved):
Looks good

hayanesuru (changes requested):
should check server is not stopped

hayanesuru (approved):

Dreeam-qwq (approved):

OverwriteMC:
100% sure this one is going to break stuff, but I cannot prove it...

PR #848, opened 2026-08-08
Optimize PoiAccess visited section tracking

HaHaWTH:
Replace LongOpenHashSet used for tracking sections visited with bit set, avoid rehash overhead.
Manually unroll the six neighbor checks, eliminate the 3x3x3 loop.

Bafore:
image

After:
image


Dreeam-qwq (approved):
tql

PR #854, opened 2026-08-09
Optimize tickPrecipitation

HaHaWTH:
1. Route tickPrecipitation biome calls to getBiomeCached
2. Pass LevelChunk into tickPrecipitation, reduce world-level lookups.


Dreeam-qwq (approved):

Dreeam-qwq:
Although there is a "world border" check in `Level#getHeight`, but I think it's fine to skip it. Doesn't matter.

hayanesuru (approved):

Dreeam-qwq:
I will add config for it later

PR #858, opened 2026-08-10
Leaves Fork cache item max stack size

HaHaWTH:
Taken from: https://github.com/Lumine1909/Leaves_Fork

Optimizes the ItemStack hotpath in item merging, inventory moving


Dreeam-qwq (approved):
tql

PR #861, opened 2026-08-11
Lithium: Cache retracting piston collision shapes

HaHaWTH:
From: https://github.com/CaffeineMC/lithium/blob/develop/common/src/main/java/net/caffeinemc/mods/lithium/mixin/block/moving_block_shapes/PistonMovingBlockEntityMixin.java

This port additionally requires the moved state's facing to match the block entity direction to preserve the original collision shape when Paper's `allow-permanent-block-break-exploits` setting creates a direction mismatch.

### DIFF
--- a/leaf-server/minecraft-patches/features/0327-Lithium-Cache-retracting-piston-collision-shapes.patch
+++ b/leaf-server/minecraft-patches/features/0327-Lithium-Cache-retracting-piston-collision-shapes.patch
@@ -1,0 +1,120 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: HaHaWTH <102713261+HaHaWTH@users.noreply.github.com>
+Date: Tue, 9 Nov 2077 00:00:00 +0800
+Subject: [PATCH] Lithium: Cache retracting piston collision shapes
+
+Precompute the 18 vanilla retracting source-piston collision shapes for
+six directions and three movement offsets.
+
+This port additionally requires the moved state's facing to match the block entity direction.
+This preserves the original collision shape when Paper's allow-permanent-block-break-exploits setting creates a direction mismatch.
+
+This patch is based on the following mixin:
+"net/caffeinemc/mods/lithium/mixin/block/moving_block_shapes/PistonMovingBlockEntityMixin.java"
+By: 2No2Name <2No2Name@web.de>
+As part of: Lithium (https://github.com/CaffeineMC/lithium)
+Licensed under: LGPL-3.0-only (https://www.gnu.org/licenses/lgpl-3.0.html)
+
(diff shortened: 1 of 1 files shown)


leaf-server/minecraft-patches/features/0327-Lithium-Cache-retracting-piston-collision-shapes.patch
```diff
@@ -0,0 +1,120 @@
...
++                );
++                mergedShapes[getIndexForMergedShape(offset, facing)] = Shapes.or(baseShape, offsetHead);
++            }
++        }
++
++        return mergedShapes;
++    }
++
++    private static int getIndexForMergedShape(final float offset, final Direction direction) {
++        if (offset != 0.0F && offset != 0.5F && offset != 1.0F) {
```
Dreeam-qwq: I think this check is useless. It's a defensive check for custom mods.
See lithium's comment "This doesn't happen in vanilla, but we fall back to vanilla code in case mods use custom offsets" under `skipVoxelShapeUnion`.

And we already have a check `extendedProgress == 0.0F || extendedProgress == 0.5F || extendedProgress == 1.0F` to avoid some strange values from `loadAdditional` write, so the check here can be removed.

leaf-server/minecraft-patches/features/0327-Lithium-Cache-retracting-piston-collision-shapes.patch
```diff
@@ -0,0 +1,120 @@
...
++        // Leaf start - Lithium - cache retracting piston collision shapes
++        final boolean retractingSourcePiston = !this.extending && this.isSourcePiston && this.movedState.getBlock() instanceof PistonBaseBlock;
++        VoxelShape pistonHeadShape = Shapes.empty();
++        // Leaf end - Lithium - cache retracting piston collision shapes
+ 
+         Direction noClipDirection = NOCLIP.get();
+         if (this.progress < 1.0 && noClipDirection == this.getMovementDirection()) {
++            // Leaf start - Lithium - cache retracting piston collision shapes
++            if (retractingSourcePiston) {
++                pistonHeadShape = this.movedState.setValue(PistonBaseBlock.EXTENDED, true).getCollisionShape(level, pos);
```
Dreeam-qwq, line 49: Should we add Lithium's shape cache? Or add it in another PR/patch?

Dreeam-qwq (changes requested):
Thanks~
Not sure whether we need to add comments from lithium too (

leaf-server/minecraft-patches/features/0327-Lithium-Cache-retracting-piston-collision-shapes.patch
HaHaWTH: I prefer adding it in another PR, this patch is only for piston 18 direction cache

Dreeam-qwq (approved):

PR #862, opened 2026-08-11
Lithium: optimize non-POI block searches

HaHaWTH:
Check patch header for details


Dreeam-qwq (approved):
Z2FtZVsuXWJpbGliaWxpWy5dY29tL2Jkb24vb25ud3l5MDEv

PR #869, opened 2026-08-14
Fix wrong baby villager sensor condition

HaHaWTH:
Fix wrong condition in `VillagerBabiesSensor`, baby checking is using the sensor owner by mistake, this will write non-villager babies into memory.


Dreeam-qwq (approved):
窝错辽

PR #871, opened 2026-08-14
fix PatrollingMonster despawn

hayanesuru:
simplify axis update

fix PatrollingMonster unable to despawn

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/world/DespawnMap.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/world/DespawnMap.java
@@ -91,7 +91,7 @@ public final class DespawnMap implements Consumer<Entity> {
 
     private boolean checkDespawn(final Entity entity) {
         EntityType<?> t = entity.getType();
-        if (entity instanceof final EnderDragon enderDragon) {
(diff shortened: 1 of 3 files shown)


leaf-server/src/main/java/org/dreeam/leaf/world/DespawnMap.java
```diff
@@ -66,20 +64,22 @@ public boolean tick(final ServerLevel world, final EntityTickList entityTickList
...
+        for (int j = 0; j < players.length; j++) {
+            final ServerPlayer p = players[j];
             if (EntitySelector.PLAYER_AFFECTS_SPAWNING.test(p)) {
-                pxl[i] = p.getX();
-                pyl[i] = p.getY();
-                pzl[i] = p.getZ();
+                playerX[i] = p.getX();
+                playerY[i] = p.getY();
+                playerZ[i] = p.getZ();
+                players[i] = p;
```
Dreeam-qwq, line 78: I think `players[i] = p;` here is useless, right?
However, it's not a big deal. approved.

Dreeam-qwq (approved):

PR #873, opened 2026-08-14
Configurable locator bar range

HyperGaming99:
Adds `gameplay-mechanisms.locator-bar-max-distance`, a cap on how far away players show up on each other's locator bar. Defaults to `0`, which keeps vanilla behaviour.

### Why

Vanilla already derives the range from the `waypoint_transmit_range` and `waypoint_receive_range` attributes, but for players both default to `6.0E7` (`Player.java`):

```java
.add(Attributes.WAYPOINT_TRANSMIT_RANGE, 6.0E7)
.add(Attributes.WAYPOINT_RECEIVE_RANGE, 6.0E7);
```

So in practice every player in a dimension is transmitted to every other one, and the connection table in `ServerWaypointManager` grows quadratically. At 100 players that is roughly 9900 waypoint connections being updated and sent.

Servers that want the locator bar as a local awareness feature rather than a dimension-wide radar currently have no way to say so short of editing the attributes through a datapack or per-player modifiers.

### Implementation

One clamp in `WaypointTransmitter#doesSourceIgnoreReceiver`:

```java
double broadcastRange = Math.min(
    source.getAttributeValue(Attributes.WAYPOINT_TRANSMIT_RANGE), receiver.getAttributeValue(Attributes.WAYPOINT_RECEIVE_RANGE)
);
// Leaf start - Configurable locator bar range - clamp only, never widen what the attributes allow
final double configuredRange = org.dreeam.leaf.config.modules.gameplay.LocatorBarRange.maxDistance;
if (configuredRange > 0.0D && configuredRange < broadcastRange) {
    broadcastRange = configuredRange;
}
// Leaf end - Configurable locator bar range
return source.distanceToSqr(receiver) >= Mth.square(broadcastRange);
```

That method is the choke point both paths already route through, so one hook covers everything:

- connection creation, via `LivingEntity#makeWaypointConnectionWith`
- `isBroken()` on all three connection types (azimuth, block, chunk)

New connections stop forming beyond the limit, and existing ones report broken so `ServerWaypointManager#updateConnection` disconnects them. Neither call site needed changing.

The value **clamps only, it never widens**. If a plugin or datapack lowers the attributes below the configured value, the attributes still win.

### What this does not do

The pair scan itself is unchanged. `ServerWaypointManager#updatePlayer` and `#updateWaypoint` still walk every player/waypoint combination; what the cap removes is the connection objects and the packets, not the distance checks. Narrowing the scan would need a spatial pre-filter and is a separate change, and I have not profiled whether it is worth the complexity.

### Testing

Built and booted against `ver/26.2`. Verified that the config entry is generated, that the compiled `WaypointTransmitter` reads `LocatorBarRange.maxDistance` before the `Mth.square` call, and that the server starts cleanly with the option set.

### DIFF
--- a/leaf-server/minecraft-patches/features/0327-Configurable-locator-bar-max-distance.patch
+++ b/leaf-server/minecraft-patches/features/0327-Configurable-locator-bar-max-distance.patch
@@ -1,7 +1,7 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: HyperGaming99 <130169800+HyperGaming99@users.noreply.github.com>
 Date: Fri, 14 Aug 2026 04:52:05 +0200
-Subject: [PATCH] Configurable locator bar range
+Subject: [PATCH] Configurable locator bar max distance
 
 Player waypoint range comes from the waypoint_transmit_range and
 waypoint_receive_range attributes, both of which default to 6.0E7 for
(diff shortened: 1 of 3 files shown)


Dreeam-qwq (approved):
Nice PR!

PR #877, opened 2026-08-15
Leaf Config refactor

### LINKED ISSUE #41: region-format configuration in purpur.yml instead of in leaf_global_config.toml
This feature is actually in the `purpur.yml` instead of `leaf_global_config.toml`
*If this works as expected, please ignore*
ce788dd3baa09b499daeb9b982da3071

Dreeam-qwq:
## Global config refactor

Refactored the Leaf config framework with clearer naming and improved scalability for the newly introduced world config impl.
The config module design is based on Luminol's config design and introduces two annotations:

- `@DoNotLoad` — Marks non-config option fields that are assigned from config values at runtime.
- `@HotReloadUnsupported` — Marks options whose values must not change during config reloads.

 Example global config module class:
 ```
 @ConfigClassInfo(category = ConfigCategory.MISC, name = "global-config-example", comments = {
     "An example global configuration section.",
     "一个全局配置节示例。"
 })
 public final class GlobalConfigExample implements ConfigModule {
 
     @ConfigInfo(name = "reloadable-value", comments = {
         "An example global value that supports hot reload.",
         "一个支持热重载的全局配置示例。"
     })
     public static String reloadableValue = "global-default";
 
     @HotReloadUnsupported
     @ConfigInfo(name = "restart-required-value", comments = {
         "An example global value that will not change when reloaded.",
         "一个不支持热重载的全局配置示例。"
     })
     public static String restartRequiredValue = "global-restart-required";
 
     @DoNotLoad
     public static String runtimeValue;
 
     @Override
     public void onLoaded() {
         runtimeValue = reloadableValue + ':' + restartRequiredValue;
     }
 }
 ```
## World config

Implemented Leaf world config.
Unlike Paper, Leaf does not automatically generate an empty config file in every world directory. This keeps world directories clean (user-friendly!). The overridden world config has any effect only when the user manually creates a `leaf-world.yml` file and adds the desired options.

The world config module is also per-file config similar to Leaf global config, and correspond field is provided in `LeafWorldConfig` for easy access.

 Example world config module class:
 ```
 @HotReloadUnsupported
 @ConfigClassInfo(category = ConfigCategory.MISC, name = "world-config-example", comments = {
     "An example world configuration section.",
     "一个世界配置节示例。"
 })
 public final class WorldConfigExample implements WorldConfigModule {
 
     @ConfigInfo(name = "restart-required-value", comments = {
         "An example world value that will not change when reloaded",
         "一个不支持热重载的世界配置示例。"
     })
     public String restartRequiredValue = "world-restart-required";
 }
```

### Leaf general config migration

The general Leafconfig migration is for migrating the option config if config paths have changed.
The migration is based on the `config-version` and supports:

- global → global
- global → world defaults
- world defaults → global
- world defaults → world defaults

So, users no longer need to manually update values when config paths are renamed.

### Gale config migration

The Gale config has been removed, and all existing Gale options have been moved into Leaf modules.
A migration process is provided to preserve existing changed values.

Migration steps:
1. Migrate Gale global and world default config when Leaf config is loaded.
2. Migrate each `gale-world.yml` to `leaf-world.yml` under each dimension directory when each level is loaded. (Config files under unloaded levels will be ignored)
3. Move all Gale config files to the `config/backup<timestamp>/` directory after the migration.
    - `gale-global.yml`
    - `gale-world-deault.yml`
    - `gale-world.yml` under each loaded dimension directory
4. Print a warning notifying the user to manually verify the migrated config.

Note: Gale world override config migration is skipped if `leaf-world.yml` already exists, `gale-world.yml` contains malformed config, or other errors occur. The step 3 and 4 will still run. Here, I assume the Gale config can load sucessfully before.

### Fixes

Fixed missing comments on reloading config


Dreeam-qwq:
For now, only the config framework is complete. The actual Leaf config module refactor and Gale migration mappings will be added later.
Feel free to provide suggestions or any issues found. (
Also, some current implementations may be a bit messy, because they are vibing by AI, and should clean up more (

There are some TODOs, can be ignored during the review:
- Comments in these PR changes
- The config and migration validations will be done once I finish all Leaf config modules and migration mapping work.

PR #878, opened 2026-08-15
Fix persistent enchanting table lapis duplication

HaHaWTH:
Purpur's implementation will dupe lapis when multiple players open the same enchanting table, this PR fixed by making EnchantingTableBlockEntity hold the shared lapis container, this will automatically sync among players


Dreeam-qwq (approved):
这个dupe这么好触发我去！！！！！

PR #886, opened 2026-08-18
Support wildcards in world name lists

### LINKED ISSUE #807: [Feature] new world list system
### What feature do you want to see added?

Hello,

I'd like to see a more advanced world list system.
In the following configuration:

```
  disable-world-data-saving:
    # Worlds listed here will skip world data persistence.
    # Changes in chunks/entities remain in memory until unload/restart and are not written to disk.
    worlds:
    - auth_lobby
    - hub01
    - hub02
    - hub03
```

I suggest that, instead of having to type **“hub01,” “hub02,”** etc., every time, we implement a “contains” system, so that, for example, typing the word **“hub”** would work for _all worlds containing the word “hub”_.

I hope this proposal is accepted and implemented quickly.
Thank you

HyperGaming99:
Closes #807.

World name lists in the config can now use `*` as a wildcard instead of spelling out every world:

```yaml
misc:
  disable-world-data-saving:
    worlds:
      - auth_lobby   # exact name, as before
      - hub*         # hub01, hub02, ...
      - '*hub*'      # every world whose name contains "hub"
```

`*` matches any (possibly empty) sequence of characters, and a pattern has to match the whole
name, so `hub*` does not match `myhub01`. Entries without `*` keep being matched exactly, so
existing configs behave exactly as before.

### Implementation

`org.dreeam.leaf.config.WorldList` holds the exact names in a `HashSet` and compiles the
wildcard entries into `Pattern`s once at config load, so matching does no per-call parsing.
`DisableWorldDataSaving.shouldSkipSave` keeps its `isEmpty()` short circuit, meaning servers
that do not use the option pay nothing. The only call sites are chunk/entity/POI saving and
`ServerLevel#save`, next to NBT serialization and disk IO, so a regex match there is noise.

Currently `disable-world-data-saving.worlds` is the only world list in the config; `WorldList`
is written so any future one can reuse it.

### Notes

The issue proposed a `--hub` prefix to mean "contains". I went with `*` instead: it expresses
contains (`*hub*`), prefix (`hub*`) and suffix (`*hub`) with the same syntax, it is the
convention users already know, and `*` is not a legal character in world folder names on
Windows, so it cannot collide with an existing exact entry. Happy to switch to another syntax
if you prefer.

Matching is case sensitive, like world folder names.

### Testing

Verified against the compiled class with assertions covering: regex metacharacters in world
names staying literal (`a+b`, `c(d)`, `e.f*`, `*g|h`), `Pattern.quote` with an embedded `\E`,
multiple and adjacent wildcards, anchoring (`hub` not matching `hub01`), unicode names,
case sensitivity, and blank/null entries being skipped.

### DIFF
--- a/leaf-server/src/main/java/org/dreeam/leaf/config/WorldList.java
+++ b/leaf-server/src/main/java/org/dreeam/leaf/config/WorldList.java
@@ -1,0 +1,72 @@
+package org.dreeam.leaf.config;
+
+import org.jspecify.annotations.NullMarked;
+
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.regex.Pattern;
+
+/// A world name list read from the config, matching either exact names or `*` wildcard patterns.
+///
+/// `*` matches any (possibly empty) sequence of characters, so `hub*` matches `hub01` and `hub02`,
+/// and `*hub*` matches every world whose name contains `hub`. Entries without `*` are matched
+/// exactly, as before. Matching is case sensitive, like world folder names.
+@NullMarked
+public final class WorldList {
+
+    private static final Pattern[] NO_PATTERNS = {};
+
+    public static final WorldList EMPTY = new WorldList(List.of());
+
+    private final Set<String> names = new HashSet<>();
+    private final Pattern[] patterns;
+
+    public WorldList(List<String> entries) {
+        List<Pattern> patterns = new ArrayList<>();
+        for (String entry : entries) {
+            if (entry == null || entry.isBlank()) {
+                continue;
+            }
+            String trimmed = entry.trim();
+            if (trimmed.indexOf('*') == -1) {
+                this.names.add(trimmed);
+            } else {
+                patterns.add(compile(trimmed));
+            }
+        }
+        this.patterns = patterns.toArray(NO_PATTERNS);
+    }
+
+    public boolean isEmpty() {
+        return this.names.isEmpty() && this.patterns.length == 0;
+    }
+
+    public boolean contains(String worldName) {
+        if (this.names.contains(worldName)) {
+            return true;
+        }
+        for (Pattern pattern : this.patterns) {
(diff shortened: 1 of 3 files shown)


leaf-server/src/main/java/org/dreeam/leaf/config/WorldList.java
```diff
@@ -0,0 +1,72 @@
...
+    private final Set<String> names = new HashSet<>();
+    private final Pattern[] patterns;
+
+    public WorldList(List<String> entries) {
+        List<Pattern> patterns = new ArrayList<>();
+        for (String entry : entries) {
+            if (entry == null || entry.isBlank()) {
+                continue;
+            }
+            String trimmed = entry.trim();
```
Dreeam-qwq: Do we need to trim here?

leaf-server/src/main/java/org/dreeam/leaf/config/WorldList.java
```diff
@@ -0,0 +1,72 @@
+package org.dreeam.leaf.config;
```
Dreeam-qwq: I think we can move this list to `org.dreeam.leaf.util.list`

leaf-server/src/main/java/org/dreeam/leaf/config/WorldList.java
HyperGaming99: No, removed. The old code matched with a plain `List#contains`, so trimming would have silently changed matching for existing configs (` hub ` suddenly matching `hub`). Entries are now used as-is; the null/blank skip stays for empty YAML list items.

leaf-server/src/main/java/org/dreeam/leaf/config/WorldList.java
HyperGaming99: Moved to `org.dreeam.leaf.util.list`. Kept `@NullMarked` on the class since that package has no `package-info.java`, unlike `org.dreeam.leaf.util`.

Dreeam-qwq (approved):

PR #899, opened 2026-08-30
bundle tracker packets so we don't spam netty

in-Piyush:
right now TrackerCtx.sendPacket was sending one connection.send per packet, so if a player had like 3 packets in the same tick it did 3 separate netty wakeups. changed it so if there's 1 packet it sends directly, if there's 2+ it sends one ClientboundBundlePacket instead. same thing we already do for startSeen, just extends it to the rest.


leaf-server/src/main/java/org/dreeam/leaf/async/tracker/TrackerCtx.java
```diff
@@ -360,11 +360,14 @@ private static void flush(ServerLevel world, Object2ObjectOpenHashMap<ServerPlay
...
-                connection.send(packetsRaw[i]);
-            }
+        if (world != connection.getPlayer().level() || list.isEmpty()) {
+            return;
+        }
+        if (list.size() == 1) {
+            connection.send(list.get(0));
+        } else {
+            // Leaf - bundle to reduce netty wakeups (ClientboundBundlePacket is vanilla, already used for startSeen)
+            connection.send(new ClientboundBundlePacket(new java.util.ArrayList<>(list)));
```
hayanesuru: the clientside has a hard limit of size on `ClientboundBundlePacket`. Netty doesn't have performance issues. It's mostly the ProtocolLib fault, though

hayanesuru (changes requested):

leaf-server/src/main/java/org/dreeam/leaf/async/tracker/TrackerCtx.java
```diff
@@ -359,12 +360,22 @@ private static void flush(ServerLevel world, Object2ObjectOpenHashMap<ServerPlay
...
+
+        if (list.size() == 1) {
+            connection.send(list.get(0));
+            return;
+        }
+
+        final int maxPackets = BundlerInfo.BUNDLE_SIZE_LIMIT;
+        for (int i = 0, size = list.size(); i < size; i += maxPackets) {
+            int end = Math.min(i + maxPackets, size);
+            connection.send(new ClientboundBundlePacket(list.subList(i, end)));
```
hayanesuru: nested BundleDelimiter should be counted (BundleDelimiter packets in BundleDelimiter packet)

hayanesuru (changes requested):

leaf-server/src/main/java/org/dreeam/leaf/async/tracker/TrackerCtx.java
```diff
@@ -359,12 +360,22 @@ private static void flush(ServerLevel world, Object2ObjectOpenHashMap<ServerPlay
...
+
+        if (list.size() == 1) {
+            connection.send(list.get(0));
+            return;
+        }
+
+        final int maxPackets = BundlerInfo.BUNDLE_SIZE_LIMIT - 2;
+        for (int i = 0, size = list.size(); i < size; i += maxPackets) {
+            int end = Math.min(i + maxPackets, size);
+            connection.send(new ClientboundBundlePacket(list.subList(i, end)));
```
hayanesuru: compilation failed

hayanesuru (changes requested):
before:
```
BundlePacket -------------------- BundlePacket  PacketB
BundleDelimiter PacketA PacketB BundleDelimiter PacketB
```

after:
```
BundlePacket ---- BundlePacket  PacketA PacketB  BundlePacket -----
BundleDelimiter BundleDelimiter PacketA PacketB BundleDelimiter PacketB BundleDelimiter
```

hayanesuru (approved):

leaf-server/src/main/java/org/dreeam/leaf/async/tracker/TrackerCtx.java
```diff
@@ -360,11 +363,38 @@ private static void flush(ServerLevel world, Object2ObjectOpenHashMap<ServerPlay
...
+            connection.send(list.get(0));
+            return;
+        }
+
+        int i = 0;
+        int j = 0;
+        while (j < list.size()) {
+            if (list.get(j) instanceof BundlePacket<?>) {
+                connection.send(new ClientboundBundlePacket((Iterable) list.subList(i, j)));
+                i = j;
```
Leymooo: infinity loop here, and btw is it really hard to use normal names instead of i, j? why do you obfuscate your own code? 

This commit itself seems to be broken, when it sees BundlePacket it combines previous packets into single bundle but why you skip BundlePacket packet from list.get(j) ?

```
connection.send(new ClientboundBundlePacket((Iterable) list.subList(i, j)));
connection.send(list.get(j)); // <-- missed??
```

and if you do this kind of optimization, cant you just unpack bundles and combine regular packets with bundled packets into one(or multiple) large bundle(s)?

in-Piyush: @hayanesuru 
made some changes i am new & i am still learning

leaf-server/src/main/java/org/dreeam/leaf/async/tracker/TrackerCtx.java
```diff
@@ -368,11 +371,39 @@ private static void flush(ServerLevel world, Object2ObjectOpenHashMap<ServerPlay
...
+        }
+
+        int end = 0;
+        int start = 0;
+        while (start < list.size()) {
+            if (list.get(start) instanceof BundlePacket<?> packet) {
+                if (end != start) {
+                    connection.send(new ClientboundBundlePacket((Iterable) list.subList(end, start)));
+                }
+                end = start;
```
Dreeam-qwq:
```suggestion
                end = start + 1;
```

Dreeam-qwq (approved):
gugugaga

PR #900, opened 2026-08-30
Optimize SpreadingSnowyBlock light retrieval

HaHaWTH:
1. Use the chunk we already get
2. Skip blockLight once skyLight >= 9
3. Short-circuit earlier at night
image


Dreeam-qwq (approved):
tql

PR #906, opened 2026-09-02
Remove final modifier from optimize PalettedContainer#get

HaHaWTH:
Some plugins may override PalettedContainer to implement features.

WHY DID THEY DO THIS?????


Dreeam-qwq (approved):
神
一会我再补一下注释

PR #907, opened 2026-09-03
Don't spam time packets with default config

HaHaWTH:


Dreeam-qwq (approved):
I'll move Purpur out once I finish my config framework refactor. Purpur breaks too many things.

PR #908, opened 2026-09-05
Cleanup optimized brain collections

Dreeam-qwq:
### 1. Rename local vars to be more standard and readable
   Replaced names like `size1`, `map1`.
   Some naming suggestions refer to the Codex; let me know which names we should keep original one or which can still be improved (

### 2. Use local vars for instance variables to fix IOOB
  Fixed https://discord.com/channels/1145991395388162119/1469315324502605854
  Use the local var to store these instance field arrays to avoid iterating over them directly. In `Brain#clearMemories`, we clear these arrays by replacing them with the empty array, so we may get IOOB during iteration, since the array is replaced.


HaHaWTH (approved):
xkl

hayanesuru (approved):

PR #913, opened 2026-09-06
Optimize EntityActivation early exit

in-Piyush:
Avoid unnecessary EntityActivation work when no players are present.

When playerSize is 0, return before building the KD-trees and scanning entities, since no entities can be activated without any affected players.

The normal activation path remains unchanged.


Dreeam-qwq (approved):
thanks~

PR #915, opened 2026-09-09
Reduce redundant seenBy update

Dreeam-qwq:
Update seenBy only once for every `handleStartTrack` (same as `handleStopTrack`), to reduce unnecessary array copy.


HaHaWTH (approved):
tql

